PR #59, opened 2016-07-28
Fixes

tomasbartkus:
First of all - good job! :) You saved me a lot of time. Have been using fake_sns for a year but it is just sooo slow, so was about to write my own implementation and then i found this repo! 

Anyways, as i said i have been using fake_sns and we have hundreds of integration tests that we use to test our service here internally. When i have tried to use this impl I had to implement these fixes.

Now these integration tests we run against fake_sns we also run against real SNS/SQS to make sure those are compatible. So with fixes in this PR our suite passes all tests.

Changes:
 - Add region to config
 - While parsing queue name we need to split it by ':' and use last element as queue name
 - Return message based on protocol, basically currently it returns json with two keys `default` and `sqs`, based on protocol we need to return only one

If you are willing to maintain this repo I will gladly contribute to it :), but if not, then please let me know.

Thank you!

### DIFF
--- a/conf/config.go
+++ b/conf/config.go
@@ -28,6 +28,7 @@ type EnvQueue struct {
 type Environment struct {
 	Host        string
 	Port        string
+	Region      string
 	LogMessages bool
 	LogFile     string
 	Topics      []EnvTopic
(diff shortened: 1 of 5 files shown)


p4tin:
@tomasbartkus - Thank you for the kind words and the PR.  I will certainly maintain this into the foreseeable future as future as we also use it for all out integration tests here.  Let me give this a run in our environment and make sure that your changes are compatible with our suite as well.  If that all goes well which I do not anticipate a problem I will merge this at the latest over the weekend.

Thanks!

tomasbartkus:
Awesome! Thanks!

PR #61, opened 2016-07-28
Add multi port to server, one for sns, one for sqs

tomasbartkus:
This PR should be merged after Godeps PR!

As this service replaces sns and sqs we need it to listen on two different ports to mimic real services. In general it works same on both ports :)

Within this PR also added "gopkg.in/check.v1" that makes testing more easier and "viper" for config loading and unmarshaling.


p4tin:
@tomasbartkus Can you explain a bit more why you need the split into 2 ports - this would be a breaking change for us and I am not too sure as to why it is needed?  If you talk directly to SQS/SNS the endpoints (i.e.: sns.us-west-2.amazonaws.com, sqs.us-west-2.amazonaws.com) are different urls and point to a farm of servers (and you could get the same server for each) but they all talk on 1 port (443).

tomasbartkus:
Point is I could make it work only with one port, but currently we have those two mocks fake_sns and fake_sqs running on two different ports in our setup. For example one of service has such configuration in `docker-compse.yaml`:
```
...
        environment:
            - GO15VENDOREXPERIMENT=1
            - AWS_ACCESS_KEY_ID=FOO
            - AWS_SECRET_ACCESS_KEY=BAR
            - AWS_DEFAULT_REGION=eu-west-1
            - AWESOME_SERVICE_DYNAMODBENDPOINT=http://dynamo.awesome:8000
            - AWESOME_SERVICE_SNSENDPOINT=http://sns.awesome:9292
            - AWESOME_SERVICE_SQSENDPOINT=http://sqs.eu-west-1:9324
```

Now, I can go and request everyone who is using my service to change this so that all would connect to `4100` instead of 9292 and 9324, but as i got 5 different services, that means 5 different teams to coordinate this :). It is not much but still a pain.

And I can image that there can be more people with such use case, so I just thought to allow them to connect on multiple ports.

Now to make it backwards compatible perhaps I could modify this as follows:

`If you will provide same port for SQS and SNS we will listen just on that one, if you will provide in config different ones we will listen on both` in this way it will work as it does now but will also allow anyone to listen on two different ports.

What do you think ?

p4tin:
I think i would prefer leaving the 'port' option ad is and add in the sqs port and and sns port.  If there I'd no port but the other 2 are pulsed then us the 2 port config.  That way we remain backwards compatible for people that have it already setup the way i have it now.

PR #65, opened 2016-07-30
Add vendors

tomasbartkus:
What about adding vendors dir ? From my experience having dependency repo disappearing is no fun :)


p4tin:
Yes I have been meaning to do that but work has been crazy - thanks!!

p4tin:
I will also look at implementing your 2 port idea in a backwards compatible way!  I might have that done by tomorrow.

PR #66, opened 2016-07-30
Add two port config for yopa and or (fake-sns + fake-sqs) compatibility

p4tin:
- Added 2 port congigurating in yaml
 - Remove the port option from the command line.

### DIFF
--- a/conf/config.go
+++ b/conf/config.go
@@ -28,6 +28,8 @@ type EnvQueue struct {
 type Environment struct {
 	Host        string
 	Port        string
+	SqsPort	    string
+	SnsPort     string
 	Region      string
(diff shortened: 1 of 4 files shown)


p4tin:
@tomasbartkus - Review and let me know if this is what you intended..  Note that in the 2 port config it does not matter which port you use for what they both fully support the entire API.

p4tin:
@tomasbartkus - let me know if this is what you intended...  Note that in the two port config both ports support the full API so they are interchangeable.

tomasbartkus:
Yep, i am preparing PR :)

p4tin:
@tomasbartkus this is the PR ;)

PR #77, opened 2016-11-25
Bugfix: SNS properly uses default message

kiril-pirozenko-home24:
If message structure is not marked as JSON, message string must be used as is for all protocols.

If message structure is marked as JSON, it must have "default" element, and it must be used if no protocol specific message was provided.


p4tin:
I'll look at this tomorrow - thanks for the PR!!

PR #109, opened 2017-06-26
PoC for lambda subscription

tomasbartkus:
Hello! (WIP - Do not merge yet!)

Wanted to share with you this PoC and ask if you will be willing to merge it once it's done :)

We need to be able to trigger lambda over SNS locally. We got a separate mini tool "lambda executor" which I want to use as aws lambda compatible API endpoint. 

Basically idea is that we implement `Invoke()` lambda API method on `lambda` SNS protocol. Once message arrives we invoke that method. 

what do you think ?

Regards!

### DIFF
--- a/app/gosns/gosns.go
+++ b/app/gosns/gosns.go
@@ -15,6 +15,10 @@ import (
 	"github.com/p4tin/goaws/app"
 	"github.com/p4tin/goaws/app/common"
 	sqs "github.com/p4tin/goaws/app/gosqs"
+	"github.com/aws/aws-sdk-go/aws"
+	"net"
+	"github.com/aws/aws-sdk-go/service/lambda"
+	"github.com/aws/aws-sdk-go/aws/session"
 )
 
 type SnsErrorType struct {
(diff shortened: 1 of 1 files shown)


p4tin:
Looks interesting, would you also be including the mini lambda executor into goaws?  Since that would now be an option if might be relevant to include it and not really call it by http call but an internal method call (using go routine???).  What do you think?

tomasbartkus:
Well this "executor" currently is a separate tool written in python so we can't really include it here.

And as lambdas are natively do not support Go, we wanted to have a separate runtime env for python and node.js. If we would execute lambdas directly for goaws we need to be able to pass over env variables (as lambdas can have env params), container of goaws would have to have python and node.js. I think it would make sense to keep it separately. And in this case we need some sort of standard API over which anyone could create their "executor" if needed (let's say for Java runtime). And I believe that AWS Lambda API [Invoke](https://docs.aws.amazon.com/sdk-for-go/api/service/lambda/#Lambda.Invoke) method does the job here. It's documented, we do not need to invent our own, so anyone can easily create endpoint which exposes it. 

We could even open one more port on goaws and implement that as well :).

Also we will release that "python lambda executor" as open source project.

p4tin:
@tomasbartkus Please re-open when you are ready but also have a look at: https://github.com/metrue/fx

PR #114, opened 2017-08-30
Add support for DeleteMessageBatch to sqs

jemmyw:
The library I'm using requires the DeleteMessageBatch api


jemmyw:
@p4tin yes, I looked into it but as there wasn't a test for `DeleteMessage` I wasn't sure where to begin, not being particularly confident in go yet. Would I be right in thinking that the current tests only test the protocol, not the underlying mechanic?

PR #123, opened 2017-12-05
Lock queues/topics until fully populated

brianmoran:


p4tin:
This is good but I should have looked at other areas of code as well...  In the CreateQueue and createTopic function in sns/gosns.go and sqs/gosqs.go there are only read-locks while i mutate the list of queues or topics - we need to change that to read/wrote locks as well.

brianmoran:
@p4tin I will look into that and update as well.

brianmoran:
@p4tin - Let me know if I missed your point. Basically I replaced the **read** locks with Lock/Unlock

app/gosqs/gosqs.go
```diff
@@ -437,7 +437,7 @@ func GetQueueUrl(w http.ResponseWriter, req *http.Request) {
 	//
 	//// Retrieve FormValues required
 	queueName := req.FormValue("QueueName")
-	SyncQueues.RLock()
+	SyncQueues.Lock()
```
p4tin: This one was not required cause the code is not changing anything just reading

app/gosqs/gosqs.go
```diff
@@ -473,7 +473,7 @@ func GetQueueAttributes(w http.ResponseWriter, req *http.Request) {
 
 	log.Println("Get Queue Attributes:", queueName)
 	if queue, ok := SyncQueues.Queues[queueName]; ok {
-		SyncQueues.RLock()
+		SyncQueues.Lock()
```
p4tin: Same comment as above

brianmoran:
@p4tin - Made the changes you suggested.

PR #132, opened 2018-01-22
Fixes #131: Build goaws for amd64 rather than arm

### LINKED ISSUE #131: Cannot start Docker container on Linux
I'm trying to run the `goaws` container (version: `latest`) on an EC2 instance like this:
```
docker run --rm -it -p 4100:4100 --name sqs -v $PWD/goaws.yml:/goaws.yml pafortin/goaws -config /goaws.yml
```

And I get the following error message:
```
standard_init_linux.go:185: exec user process caused "exec format error"
```

queeno:
The amd64 instruction set is compatible with the vast majority of CPUs.
Let's build amd64 executables rather than arm.

### DIFF
--- a/makefile.txt
+++ b/makefile.txt
@@ -19,7 +19,7 @@ git-release:
 	curl --data $(APIJSON) https://api.github.com/repos/p4tin/goaws/releases?access_token=$(GITHUB_API_KEY)
 
 linux:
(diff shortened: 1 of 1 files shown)


queeno:
Hi @p4tin 
Could you please also generate and upload a new container to dockerhub?
Thanks!

p4tin:
The new docker was generated!

queeno:
Thank you! 👍

PR #139, opened 2018-01-26
add attributes support to ReceiveMessage

jeloou:
This PR adds attributes to the output format of the `ReceiveMessage` action. It also refactors `app/gosqs/sosqs.go` to make it easier to read.

### DIFF
--- a/app/sqs.go
+++ b/app/sqs.go
@@ -3,8 +3,6 @@ package app
 import (
 	"sync"
 	"time"
-
-	"github.com/p4tin/goaws/app/common"
 )
 
 type SqsErrorType struct {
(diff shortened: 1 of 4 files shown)


app/sqs.go
```diff
@@ -3,6 +3,8 @@ package app
 import (
 	"sync"
 	"time"
+
+	"github.com/p4tin/goaws/app/common"
```
p4tin: The base app package should never import a dependancy from this project as it is in place to avoid circular dependacies - also no code goes in any of the 'app' package files - they are strictly for struct and interface declarations.

app/sqs.go
```diff
@@ -21,6 +23,48 @@ type Message struct {
 	MD5OfMessageBody       string
 	ReceiptHandle          string
 	ReceiptTime            time.Time
+	MessageAttributes      map[string]MessageAttributeValue
+}
+
+func (m *Message) GetResult() *ResultMessage {
```
p4tin: Once refactored your new functions will require unit tests.

app/sqs.go
```diff
@@ -21,6 +23,48 @@ type Message struct {
...
+}
+
+type MessageAttributeValue struct {
+	Name     string
+	DataType string
+	Value    string
+	ValueKey string
+}
+
+func (a *MessageAttributeValue) GetResult() *ResultMessageAttribute {
```
p4tin: Once refactored your new functions will require unit tests.

app/sqs.go
jeloou: @p4tin - I moved the methods code to `app/gosqs/gosqs.go`, but you should consider adding methods to the structs in `app`. 

The pattern you're following is missing the point of having structs you can attach methods to.

jeloou:
@p4tin - can you take a look at the fixes? I would be nice to have this merged.

p4tin:
@jeloou - Good Work

PR #144, opened 2018-02-12
Add ChangeMessageVisibility endpoint and implement VisibilityTimeout logic

### LINKED ISSUE #85: implement changeMessageVisibility

canthefason:
This PR introduces ChangeMessageVisibility endpoint and requeueing logic behind VisibilityTimeout. Mutex locking is becoming too strict, which needs to be improved later on.

I've also added DeadLetter Queue support. 

The PR also includes the changes from #143 and closes #85

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 [![Build Status](https://travis-ci.org/p4tin/goaws.svg?branch=master)](https://travis-ci.org/p4tin/goaws)
 
 You are always welcome to [tweet me](https://twitter.com/gocodecloud) or [buy me a coffee](https://www.paypal.me/p4tin)
- 
+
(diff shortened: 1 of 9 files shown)


p4tin:
@canthefason I agree that the mutext locking is becoming too strict - I was going to move some of the Mutext locks at the message level like for this change you made which I truly appreciate. the changing of message visibility needs to move to the message level it is changed.  But for this we need to create some methods on the message itself then lock that message when things change with that message this should improve performance and isolate the locks to where they need to be.   Sorry for being so long winded.   Thanks for your PRs it's great..

canthefason:
I've also started adding Mutex locks at the message level, then stopped in favor of limiting the scope for this PR. Thank you so much for your efforts in putting this repo together. It's been very helpful lately.

PR #148, opened 2018-02-19
Resolving deadlock issue with `WaitTimeSeconds` parameter

archa347:
The function `numberOfHiddenMessageInQueue` also tries to acquire a lock with `app.SyncQueues.Lock()`, thus using the `WaitTimeSeconds` immediately causes deadlock as it tries to acquire the lock twice.  This attempts to resolve this issue.

### DIFF
--- a/app/cmd/goaws.go
+++ b/app/cmd/goaws.go
@@ -8,9 +8,9 @@ import (
 
 	log "github.com/sirupsen/logrus"
 
-	"github.com/p4tin/goaws/app/conf"
-	"github.com/p4tin/goaws/app/gosqs"
-	"github.com/p4tin/goaws/app/router"
+	"github.com/archa347/goaws/app/conf"
+	"github.com/archa347/goaws/app/gosqs"
+	"github.com/archa347/goaws/app/router"
 )
 
 func main() {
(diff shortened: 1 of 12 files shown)


iambbs:
I have also noticed this and fixed in PR #146

p4tin:
@archa347 @iambbs I will check both PRs but as they are pretty big give me a few days as I am swapped with work at the moment....  Thank you both for the PRs - I am happy that goaws is helping you out!!

archa347:
I'm going to close mine for now, I had to make some changes to the `imports` paths to get this to build properly from my fork.   Don't really want that committed.  Is there a way to make some of those package references relative so they don't need to know whose repo they are in?

iambbs:
Thank you @p4tin for your time and this repo.
@archa347 I don’t know if there is a solution, but what I did is originally cloned p4tin’s and then just changed the origin to my fork

archa347:
@iambbs Yeah, looking around it appears this is the common solution, or
just cloning your fork into your $GOPATH manually (instead of using `go
get`) and matching the directory names to the package names in the code.

On Mon, Feb 19, 2018 at 2:28 PM, Boris Khutoretsky <notifications@github.com

PR #152, opened 2018-03-09
Fix on a deadlock for ReceiveMessage when wait time is specified

### LINKED ISSUE #151: Deadlock on sqs when calling receive-message with WaitTime greater than 0
Deadlock happens because

https://github.com/p4tin/goaws/blob/3c664cdb897460d91593499d4f07f90a0e3a9d25/app/gosqs/gosqs.go#L339

Method locks the sqs queue, calls the numberOfHiddenMessagesInQueue which also tries to lock the queue (and never gets the lock because it's already locked)

Introduced by https://github.com/p4tin/goaws/commit/3c664cdb897460d91593499d4f07f90a0e3a9d25

eligijusvitkauskas-home24:
Fixes #151

### DIFF
--- a/app/gosqs/gosqs.go
+++ b/app/gosqs/gosqs.go
@@ -335,9 +335,9 @@ func ReceiveMessage(w http.ResponseWriter, req *http.Request) {
 
 	loops := waitTimeSeconds * 10
 	for loops > 0 {
-		app.SyncQueues.Lock()
+		app.SyncQueues.RLock()
 		found := len(app.SyncQueues.Queues[queueName].Messages)-numberOfHiddenMessagesInQueue(*app.SyncQueues.Queues[queueName]) != 0
-		app.SyncQueues.Unlock()
+		app.SyncQueues.RUnlock()
 		if !found {
 			time.Sleep(100 * time.Millisecond)
 			loops--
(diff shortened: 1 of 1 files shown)


app/gosqs/gosqs.go
```diff
@@ -335,9 +335,9 @@ func ReceiveMessage(w http.ResponseWriter, req *http.Request) {
 
 	loops := waitTimeSeconds * 10
 	for loops > 0 {
-		app.SyncQueues.Lock()
+		app.SyncQueues.RLock()
```
p4tin: Now that I look at this I do not think we need a Lock/unlock around that line at all.  Just remove it.

eligijusvitkauskas-home24: But in case of removal of locks here we still keep locks in `numberOfHiddenMessagesInQueue` function then?

eligijusvitkauskas-home24: @p4tin ping

Also wouldn't you like consistent app.SyncQueues.Queues[queueName] between `len` and `numberOfHiddenMessagesInQueue` calls? Having a lock just inside numberOfHiddenMessagesInQueue wouldn't guarantee that

p4tin: you are correct

Issue #150, opened 2018-03-09
Queues remain locked in case queue is not found


mariusgrigaitis:
Reading through the code:

https://github.com/p4tin/goaws/blob/1f0ae7c001aa4ea286dc638885be4bdeef954886/app/gosqs/gosqs.go#L435

It seems like there's a bug there for not releasing queues lock which might cause queues to lock up

p4tin:
Good catch do you want to make a PR?

PR #153, opened 2018-03-09
Unlock SyncQueue before early return

eligijusvitkauskas-home24:
Fix for #150


app/gosqs/gosqs.go
```diff
@@ -432,6 +432,7 @@ func ChangeMessageVisibility(w http.ResponseWriter, req *http.Request) {
 	app.SyncQueues.Lock()
```
p4tin: Just move the lock to below the check then no need to Unlock it!!!

PR #157, opened 2018-03-23
FIFO Support

tomasbartkus:
This is backward compatible PR to support FIFO queues. If queue name has `.fifo` suffix it will be treated as FIFO. 

- receive a message from the same groupId will lock it and won't allow other consumers to receive it
- visibility timeout and removing message will unlock group

### DIFF
--- a/app/conf/config.go
+++ b/app/conf/config.go
@@ -62,7 +62,15 @@ func LoadYamlConfig(filename string, env string) []string {
 	for _, queue := range envs[env].Queues {
 		queueUrl := "http://" + app.CurrentEnvironment.Host + ":" + app.CurrentEnvironment.Port + "/queue/" + queue.Name
 		queueArn := "arn:aws:sqs:" + app.CurrentEnvironment.Region + ":000000000000:" + queue.Name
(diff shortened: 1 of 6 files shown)


p4tin:
@tomasbartkus You need to rebase your branch because of other PRs that I have merged sorry...  As well I'm not sure I like using 'magic' name (.fifo) for deciding this is a FIFO queue or not...  I'd like you to consider that a FIFO queue just has a flag on/off in the queue struct instead.  Let me know what you think...

tomasbartkus:
There is no magic here :) (https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/FIFO-queues.html)

```
The name of a FIFO queue must end with the .fifo suffix. The suffix counts towards the 80-character queue name limit. To determine whether a queue is FIFO, you can check whether the queue name ends with the suffix.
```

So this `.fifo` suffix acts as a toggle :)

p4tin:
@tomasbartkus Sorry for my ignorance - I'll merge your PR as soon as the conflicts are resolved!!

tomasbartkus:
@p4tin ping :)

PR #159, opened 2018-04-13
Add SNS filter policy support

forrest-ua:
This PR adds support of simple filter policy to SNS subscription. Only simple exact string match is supported at the moment, but can be extended in the future.


forrest-ua:
@p4tin could you please check this PR?

p4tin:
@forrest-ua Hey I merged the PR but I was wondering if you could also update the postman collection to show the use of the sns filter policy. Thanks!

forrest-ua:
Hi @p4tin, good point! I've imported your collection and added two 2 requests:
- GetSubscriptionAttributes
- SetSubscriptionAttributes (FilterPolicy)

So you can maybe update your link with these new requests?

PR #161, opened 2018-04-24
Attaches attributes on raw

qhenkart:
the emulator currently ignores any message attributes, even when the subscription is raw. AWS SNS however includes message attributes when the subscription is raw. https://docs.aws.amazon.com/sns/latest/dg/SNSMessageAttributes.html

with this PR, publishing with message attributes and processing those attributes in the queue now work seamlessly with the AWS-SDK for go


qhenkart:
closing this PR, because I think it would be better to add it on top of https://github.com/p4tin/goaws/pull/159 once it is accepted

PR #172, opened 2018-08-18
SNS, SQS signature check fixes - for AWS SDK

andreyserdjuk:
SNS: previous signature genaration msg.MD5OfMessageAttributes = common.GetMD5Hash("GoAws")  caused failing of signature check in AWS SDK so could be simply deleted but I've added signature gen from gosqs.
  
SQS: add MD5OfMessageAttributes only if MessageAttributes not empty - according to logic of AWS SDK signature check failed if MessageAttributes filled and MD5OfMessageAttributes exists.

### DIFF
--- a/app/gosns/gosns.go
+++ b/app/gosns/gosns.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/p4tin/goaws/app"
 	"github.com/p4tin/goaws/app/common"
+	"github.com/p4tin/goaws/app/gosqs"
 	log "github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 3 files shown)


app/gosns/gosns.go
```diff
@@ -328,10 +329,10 @@ func Publish(w http.ResponseWriter, req *http.Request) {
 						msg.MessageBody = m
 					} else {
 						msg.MessageAttributes = messageAttributes
+						msg.MD5OfMessageAttributes = gosqs.HashAttributes(messageAttributes)
```
p4tin: HashAttributes should be moved to a common package and imported from gosqs and gosns to avoid any possibilities in the future of circular reference.  Any unit tests should be moved as well.

PR #178, opened 2018-10-17
Adding attributes in received message response

mwaaas:
Sqs receive message does not return Attributes this is to add the attributes to the response

### DIFF
--- a/app/gosqs/gosqs.go
+++ b/app/gosqs/gosqs.go
@@ -843,9 +843,22 @@ func SetQueueAttributes(w http.ResponseWriter, req *http.Request) {
 }
 
 func getMessageResult(m *app.Message) *app.ResultMessage {
-	attrs := []*app.ResultMessageAttribute{}
+	msgMttrs := []*app.ResultMessageAttribute{}
 	for _, attr := range m.MessageAttributes {
-		attrs = append(attrs, getMessageAttributeResult(&attr))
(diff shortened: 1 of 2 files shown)


app/gosqs/gosqs.go
```diff
@@ -843,9 +843,22 @@ func SetQueueAttributes(w http.ResponseWriter, req *http.Request) {
...
 
 func getMessageResult(m *app.Message) *app.ResultMessage {
-	attrs := []*app.ResultMessageAttribute{}
+	msgMttrs := []*app.ResultMessageAttribute{}
 	for _, attr := range m.MessageAttributes {
-		attrs = append(attrs, getMessageAttributeResult(&attr))
+		msgMttrs = append(msgMttrs, getMessageAttributeResult(&attr))
+	}
+
+	attrsMap := map[string]string{
```
p4tin: what is the attributes "SenderId" and  ApproximateReceiveCount do seems you are settings them to "fake" values and then using those as attributes??

p4tin: I do not see also in the docs where there is a distinction between "attributes' and "messageattributes"??

jairovadillo:
@p4tin this is from the 17 oct maybe could be move forward to master? For me is pretty important to have access to the MessageAttributes from the response and this is causing discrepancies between goaws and the real AWS

p4tin:
@jairovadillo I am not sure this will solve your problem so I am going to test and see how I can help...

PR #192, opened 2019-02-08
Enable redrive policy in config file

pswaminathan:
Currently there is no support for putting a redrive policy in the config file, e.g.

```yaml
  Queues:
    - Name: failed-messages
    - Name: messages
      RedrivePolicy: '{"deadLetterTargetArn":"arn:aws:sqs:us-west-2:100010001000:failed-messages","maxReceiveCount":10}'
      VisibilityTimeout: 5
```

This adds support for it by pulling the redrive policy logic into a separate function and calling it when creating queues from the config.

### DIFF
--- a/app/common.go
+++ b/app/common.go
@@ -15,6 +15,7 @@ type EnvTopic struct {
 type EnvQueue struct {
 	Name                          string
 	ReceiveMessageWaitTimeSeconds int
+	RedrivePolicy                 string
 }
 
 type EnvQueueAttributes struct {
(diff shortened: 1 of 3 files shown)


pswaminathan:
Hi @p4tin — do you have any thoughts on this?

p4tin:
Your PR has errors in the build!

`package math/bits: unrecognized import path "math/bits" (import path does not begin with hostname)`

p4tin (changes requested):
Fix build errors!

p4tin:
I have merged #235 which provides this capability.

pswaminathan:
It is worth noting that:

1. This built and ran two years ago when I made the PR—I built off this branch and used at a previous company
2. The specific error implies an error in the Go environment (given that math/bits is in the standard library), rather than an issue with this code
3. math/bits wasn't even involved in this change
4. I hadn't rebased on master since I thought this was abandoned

I am glad the functionality is merged, in some form or fashion.

PR #198, opened 2019-08-28
Added random latency to SQS receives.

andykuszyk:
Latency is based on the inclusion of two environment variables
(`GOAWS_RANDOM_LATENCY_MIN` and `GOAWS_RANDOM_LATENCY_MAX`), which, if set
will prevent messages from being released from the queue until after a
period of random, simulated latency has elapsed.

The purpose of this change is to aid in local testing against AWS when
SQS latency is expected and needs to be developed against.


app/sqs.go
```diff
@@ -32,6 +37,43 @@ type Message struct {
...
+	randomLatency, err := getRandomLatency()
+	if err != nil {
+		log.Error(err)
+		return true
+	}
+	return m.SentTime.Add(randomLatency).Before(time.Now())
+}
+
+func getRandomLatency() (time.Duration, error){
+	minVar := os.Getenv("GOAWS_RANDOM_LATENCY_MIN")
```
p4tin: Sorry this took so long to evaluate but the values should be in the goaws.conf not in environment variables.  This is for consistency sake with how all configs are done.  thanks!

andykuszyk: @p4tin no worries, I've just pushed another commit - is this what you had in mind?

andykuszyk:
@p4tin does this PR look ok now?

PR #213, opened 2020-01-05
Get resource url from host header

AndrewChubatiuk:
Hi @p4tin 
First of all want to thank you for your project, it's really helpful.
I'm trying to spin up goaws using docker-compose with dynamic host port mapping and I had problems trying to send a message to URL that was retrieved from GetQueueUrl, cause queue url host and port are based on configuration file instead of HTTP Host header value. Also it's hard to query goaws instance from host and overlay networks at the same time.
I've submitted this PR with a resolution of this problem.

### DIFF
--- a/app/conf/config.go
+++ b/app/conf/config.go
@@ -67,19 +67,12 @@ func LoadYamlConfig(filename string, env string) []string {
 		app.CurrentEnvironment.AccountID = "queue"
 	}
 
-	if app.CurrentEnvironment.Host == "" {
-		app.CurrentEnvironment.Host = "localhost"
-		app.CurrentEnvironment.Port = "4100"
-	}
-
 	app.SyncQueues.Lock()
 	app.SyncTopics.Lock()
 	for _, queue := range envs[env].Queues {
-		queueUrl := "http://" + app.CurrentEnvironment.Host + ":" + app.CurrentEnvironment.Port +
-			"/" + app.CurrentEnvironment.AccountID + "/" + queue.Name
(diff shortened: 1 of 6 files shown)


nhoughto:
Might want to incorporate the config flag as per my fix for your same issue (i saw this after i finished! 🙄 ), not changing the default behaviour likely makes it easier to merge the PR? 🤷‍♂ 

https://github.com/bsycorp/goaws/commit/2f3217691c2e21367175ecf34c06a9a31cc6a0ae

Admiral-Piett:
Posting this here too just in case:

Following up here @nhoughto @AndrewChubatiuk - would you mind speaking a little more to my question here around the use case on this?

https://github.com/Admiral-Piett/goaws/pull/234#issuecomment-1495215750

Admiral-Piett:
Referring to comment: https://github.com/Admiral-Piett/goaws/pull/234#issuecomment-1537443027

Sorry again @AndrewChubatiuk for the delay, I didn't see this until now.  I'm circling back on this to keep it on my radar.  So the issue is that you're wanting to call a single instance of goaws from the same host it's on and from a service in a docker network - and you want the responding urls to match the network you're calling from?  AKA `localhost:4100/...` for local, and `goaws.com/...` for your docker calls?

If so, what's the problem you're having with it NOT doing this?  Assuming I'm understanding what the goal is here - it seems odd for goaws to report that it's host based on what the requester is sending in.

Admiral-Piett:
Circling back here @AndrewChubatiuk to see if you've had a chance to think about this.

PR #218, opened 2020-02-20
fix routes order

DayS1eeper:
Requests to `SigningCertURL` are currently handled by `actionHandler` instead of `pemHandler` because of wrong routes order. This PR fix this issue.


DayS1eeper:
> Nice catch - Would you mind writing a unit test to test these routes?

I have never written in GO before XD, but I'll try.

DayS1eeper:
@p4tin Could you please update the image on  Docker Hub?

PR #221, opened 2020-03-13
fix router order

kcajmagic:
`SimpleNotificationService/{id}.pem` endpoint was falling into `/` path causing the consumer to receive a Bad Request response.

### DIFF
--- a/app/router/router.go
+++ b/app/router/router.go
@@ -17,11 +17,11 @@ import (
 func New() http.Handler {
 	r := mux.NewRouter()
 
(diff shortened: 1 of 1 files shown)


DayS1eeper:
`SimpleNotificationService/{id}.pem` was caught by `/{account}/{queueName}` route and it was fixed in this PR https://github.com/p4tin/goaws/pull/218 .

kcajmagic:
My bad I was looking at v0.3.1

PR #222, opened 2020-03-13
improve docker image creation

kcajmagic:
This PR changes a couple of things.

- sets the running user to nobody so the container does not run as root. (security precaution)
- changes the base image to scratch. Makes the overall image smaller. (can make life harder to debug though)
- builds the image in docker then copies the binary over.


p4tin:
As I am currently using goreleaser to create the binaries and the docker image I'd rather stick with that process - as well I have never tested the USER option I might include that in a future release.

PR #227, opened 2020-08-30
Default configuration can never load following documentatation

qhenkart:
Currently, following the documentation, the default configuration will never load and will never cast an error when it fails.

Additionally when loading a custom configuration file, an error is never fired if there is an error in loading due to an invalid or incorrect path

this PR adds an error when the provided configuration file does not load, it also fixes the broken path that causes the default file to never load. It is only a partial fix because if someone moves the compiled binary outside of the root path, the default file will no longer be found (but at least it returns an error)

I also fixed linting errors


p4tin:
Seems like the pr-tests failed can you check and fix please?

p4tin (changes requested):
Build fails!

dhumphreys01:
I just merged another PR to handle this issue and fixed up the release process to do the same thing.  All versions should pull the default properly now I think.

Let me know if you have any trouble with versions v0.4.0 or greater.  Closing this as a dupe.

PR #228, opened 2020-09-22
Respect GetQueueAttributes AttributeName parameters

mbklein:
This PR allows `goawssqs` to respect the [GetQueueAttributes](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_GetQueueAttributes.html) `AtrributeName.N` parameter and only return requested attributes. It also removes a redundant `app.SyncQueues.RLock()` call that was causing the SQS handler to lock up indefinitely when bombarded with `GetQueueAttributes` calls.

### DIFF
--- a/app/gosqs/gosqs.go
+++ b/app/gosqs/gosqs.go
@@ -774,6 +774,27 @@ func GetQueueAttributes(w http.ResponseWriter, req *http.Request) {
 	// Retrieve FormValues required
 	queueUrl := getQueueFromPath(req.FormValue("QueueUrl"), req.URL.String())
 
+	attribute_names := map[string]bool {}
+
+	for field, value := range req.Form {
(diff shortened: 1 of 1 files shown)


p4tin (changes requested):
No Unit tests!

p4tin (dismissed):
Perhaps some unit tests?

mbklein:
Sorry, missed this. I'll get on it.

Admiral-Piett:
@mbklein Hey there!  Checking this out - looks great, would you mind doing a quick rebase and throwing some tests on it?

mbklein:
@Admiral-Piett It only took me 2½ years, but I've added some tests. Sorry this fell off my radar for so long.

Admiral-Piett (approved):
Brilliant!  You're awesome, thank you!

PR #235, opened 2021-01-20
sqs: allow redrive-policy to be set from config yaml

sebcante:
Allow to set queue `RedrivePolicy` in `config.yml`

```
    - Name: local-queue3                # Queue name
      RedrivePolicy: '{"maxReceiveCount": 100, "deadLetterTargetArn":"arn:aws:sqs:us-east-1:000000000000:local-queue3-dlq"}'  
    - Name: local-queue3-dlq  
```

### DIFF
--- a/app/conf/config.go
+++ b/app/conf/config.go
@@ -2,8 +2,10 @@ package conf
 
 import (
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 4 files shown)


app/conf/config.go
```diff
@@ -165,3 +180,40 @@ func createSqsSubscription(configSubscription app.EnvSubsciption, topicArn strin
 	newSub.SubscriptionArn = subArn
 	return newSub
 }
+
+func setQueueRedrivePolicy(queues map[string]*app.Queue, q *app.Queue, strRedrivePolicy string) error {
+	// support both int and string maxReceiveCount (Amazon clients use string)
```
sebcante, line 185: inspired from [gosqs](https://github.com/p4tin/goaws/blob/master/app/gosqs/queue_attributes.go#L32) but decided to not introduce a dependency from `conf` to `gosqs` package but open for suggestion there.

sebcante:
😮  i just realised (late) that @pswaminathan did a similar PR #192  about it. I am happy with either implementation.

 @p4tin what do you you think, would love to have that feature merged as it helps us a lot doing integration testing with Dead Letter queues

p4tin (approved):

PR #238, opened 2021-06-21
Fix xml response for list-subscriptions-by-topic

sweatybridge:
XML response format is malformed for `list-subscriptions-by-topic`. Turns out the struct was defined but unused.

Reproduced by aws cli version 2.2.13 on macOS
```
aws --endpoint-url http://localhost:4100 --debug sns list-subscriptions-by-topic --topic-arn <topic_arn>
```


Admiral-Piett (approved):
Looks great, thanks!

PR #241, opened 2021-11-25
app/gosqs: increment number of receives for a message

vishen:
When returning an SQS message, increment the 'NumberOfReceives' so it
always contains the number of times a message has been received but not
yet deleted.

From the AWS SQS docs:

    ApproximateReceiveCount – Returns the number of times a message has been received across all queues but not deleted.

### DIFF
--- a/app/gosqs/gosqs.go
+++ b/app/gosqs/gosqs.go
@@ -180,8 +180,8 @@ func SendMessage(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
-	if (app.SyncQueues.Queues[queueName].MaximumMessageSize > 0 &&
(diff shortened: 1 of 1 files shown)


Admiral-Piett:
@vishen Sorry it's been such a long time - this looks great to me.  Mind throwing a test in here for it?

Admiral-Piett:
@vishen Following up here - if you don't want to mess with this anymore (understandable) I'll branch off this, write some tests and merge that instead.  Let me know!  I appreciate the effort!

vishen:
Apologies @Admiral-Piett , I don't have the time anymore to add the tests. If you have time to add them that would be amazing :)

PR #244, opened 2022-07-11
Update permissible SNS endpoint call response codes

nicolas-s-miller:
https://docs.aws.amazon.com/sns/latest/dg/SendMessageToHttp.prepare.html

When an SNS subscription calls its endpoint, Amazon recognizes responses in the range of 200-499 as successful Notification deliveries. Responses outside of that range will trigger the Subscription's retry policy.

This PR updates the `callEndpoint` method to allow this range instead of just response codes divisible by 200.

### DIFF
--- a/app/gosns/gosns.go
+++ b/app/gosns/gosns.go
@@ -651,14 +651,19 @@ func callEndpoint(endpoint string, subArn string, msg app.SNSMessage, raw bool) 
 	if res == nil {
 		return errors.New("response is nil")
 	}
-	if res.StatusCode%200 != 0 {
+
+	//Amazon considers a Notification delivery attempt successful if the endpoint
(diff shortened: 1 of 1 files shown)


YixingCheng:
Wanted to make the same change.

Admiral-Piett (approved):
Good catch!  Looks good to me, thanks @nicolas-s-miller.

nicolas-s-miller:
I don’t have `write` access to merge this. What happens next? No rush. I’m just curious.

dhumphreys01:
Hm, still figuring out how all this is set up.  I'll take a look at the settings for the future but of course, I merged it for now!  Thanks again for having a look at this!

fffinkel:
This merge broke the app. Submitted #256 with the fix.

PR #249, opened 2023-02-24
Fix package references from repo migration

dhumphreys01:
We broke all the package references when I inherited this repo.  So this should update them so we can at least clone/build/run this repo locally.  (Docker not included - still deciding where to host it).


README.md
```diff
@@ -1,7 +1,7 @@
 # GoAws
 [![Build Status](https://travis-ci.org/p4tin/goaws.svg?branch=master)](https://travis-ci.org/p4tin/goaws)
 
-You are always welcome to [tweet me](https://twitter.com/gocodecloud) or [buy me a coffee](https://www.paypal.me/p4tin)
+You are always welcome to [tweet the creator in chief](https://twitter.com/gocodecloud) or [buy him a coffee](https://www.paypal.me/p4tin)
```
p4tin, line 4: You can just remove that or put your name there - lol

go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/p4tin/goaws
+module github.com/Admiral-Piett/goaws
 
 go 1.12
```
p4tin, line 3: Probably need to update this as well

p4tin (approved):

README.md
Admiral-Piett: Hahaha I'll leave it up for you for a bit - you have been doing the hard work here for a long and I'm not set up!

go.mod
Admiral-Piett: Yep, I was going to do that shortly.  Want to use this to get us back in working order, then start upgrading.

PR #250, opened 2023-02-24
Fix automatic config import

dhumphreys01:
The dot here references the location of the binary (or the working directory being used when running it), the binary is usually located outside the `app` folder, so we need to make sure we can traverse the whole project structure to get to the default properly.

Block by: https://github.com/Admiral-Piett/goaws/pull/249


p4tin (approved):
I put a couple of minor comments on the other PR that might apply here if you intend on closing the other one and just merging this one.

PR #255, opened 2023-03-02
Fix default config in docker release

Admiral-Piett:


williamdes:
I think you forgot this one:
https://github.com/Admiral-Piett/goaws/blob/bc027e4ec99c700f0935c75c9a57f6cc21157cb2/docker-compose.yml#L9

williamdes:
Maybe also https://github.com/Admiral-Piett/goaws/blob/bc027e4ec99c700f0935c75c9a57f6cc21157cb2/Dockerfile#L19

It's a bit unclear where the config mount is.
While using the new Docker image I had to change the mount to the /app/config path

Issue #260, opened 2023-04-24
Feature Request: Configuration Hot-Reload


apavanello:
Dear development team,
The feature I am proposing is Configuration Hot-Reload.Currently, when we need to make changes to the configuration file, we have to interrupt the entire program and restart it. This is a big inconvenience because we lost all "in-memory" data

Configuration Hot-Reload will allow us to update the configuration file without the need to restart the entire program. Additionally, the update will only affect new items or modifications in the configuration file, while exclusions will remain unchanged.

I added this feature that will create a reader for events involving the configuration file. With this, it is possible to edit the file and maintain persistence without the need to restart the program. I also added a "hack" to make it work on k8s, as it does not generate a write event in the file, but a remove event instead. The code for this feature is in the PR below, and I also added a test for it.

To activate this feature, which is disabled by default, it is necessary to start goaws with the --hot-reload flag.

Thank you for your attention.

Best regards,
Alexandre Pavanello.

Admiral-Piett:
Nice work on this one!  I'll look into getting your commits into the next release - thanks for taking care of that!

PR #261, opened 2023-04-24
add HOT-RELOAD feature and tests

apavanello:
#260

### DIFF
--- a/app/conf/config.go
+++ b/app/conf/config.go
@@ -3,7 +3,9 @@ package conf
 import (
 	"encoding/json"
 	"fmt"
+
 	"io/ioutil"
+	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -13,12 +15,14 @@ import (
 
 	"github.com/Admiral-Piett/goaws/app"
 	"github.com/Admiral-Piett/goaws/app/common"
+	"github.com/fsnotify/fsnotify"
 	"github.com/ghodss/yaml"
 )
 
 var envs map[string]app.Environment
 
 func LoadYamlConfig(filename string, env string) []string {
+
 	ports := []string{"4100"}
 
 	if filename == "" {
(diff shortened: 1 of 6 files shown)


Admiral-Piett:
@apavanello Nice, this looks awesome.  Just to clarify here - all you're enabling for Hot Reload is the queues and the topics themselves, correct?  The other things, Log File/Region/QueueAttributes/etc. those look omitted from `StartWatcher`.

Admiral-Piett:
@apavanello This change looks good, but it looks like it needs a rebase now.  I have been playing with it today and it looks like this conflicts with some of the tests.  I think they're treading on each other's memory addresses.

Anyway - I have a branch locally with some refactoring.  I'll push it when I can if you want to see - though it will be significant, not all particularly related to your code.  Thought, if you could share your k8 use case that would be really helpful - I want to make sure whatever I come up with matches your needs too.  

In the meantime, if you want to rebase this and get it working that works too and I'll merge it.  Up to you!  Let me know!

app/conf/config.go
```diff
@@ -79,8 +83,144 @@ func LoadYamlConfig(filename string, env string) []string {
...
+				}
+
+				err = yaml.Unmarshal(yamlFile, &envs)
+				if err != nil {
+					log.Errorf("err: %s", err)
+					return
+				}
+
+				log.Infoln("Load new SQS config:")
+				err = createSqsQueues(env)
```
Admiral-Piett, line 155: Also, I had a thought today - do we want to wipe the old Queues and Topics before reload the new data into them?  I figure, if something was removed, we probably want to yank that on a hot reload?  Maybe it's as easy as overwriting the queues with an empty map before repopulating?

app/conf/config.go
```diff
@@ -79,8 +83,144 @@ func LoadYamlConfig(filename string, env string) []string {
...
+
+	// Start listening for events.
+	go func() {
+		for {
+			select {
+			case event, ok := <-watcher.Events:
+				if !ok {
+					return
+				}
+				if event.Has(fsnotify.Remove) {
```
Admiral-Piett, line 122: After looking at this a bit deeper, I don't think I fully understand this block.  Is it because in your environment - this "remove" event is what happens on a file change (in a "remove" then "recreate" pattern)?  If so, the file name isn't changing is it?  Could we not just do the wait/loop to check if it's rebuilt and then proceed down with the rest of the reloading calls?

I think I'm confused about 2 things really.
1. Why call `StartWatcher` again, and synchronously?
2. Why block the outer function with the `quit` channel?  In most of these non-k8 with the remove event cases would that ever actually hang up or is that goroutine just orphaned now?

app/conf/config.go
Admiral-Piett: Further testing shows that it does this already.

apavanello:
Hello there.
Sorry, I've been working on a new project at work for the past few months, and it's been taking up a lot of my time.
I need to look at the code again because I don't remember some of the details completely. I'll try to do it tonight when I'm done working.

But I can tell you that in the case of `fsnotify.Remove`, it's for if you want to use goaws in a Kubernetes environment. Kubernetes  doesn't send clear signals of file changes or creation, it only sends a remove when the file state changes in a `configmap`.

Admiral-Piett:
Ah, I can totally understand that alternate project business, story of my life.  😀  I have a sizeable PR that I'll probably want to incorporate anyway, but I'll incorporate whatever you come up with too.  Alternately you can review my code when it's ready.  I'm fleshing out some tests but I think it's nearly ready.  I'll play with the k8 use case to make sure I'm capturing it to.  Thanks for the info!

Admiral-Piett:
@apavanello So - I am throwing this out there as a new PR.  The more I got into it, the more I realized it's a major feature. but my PR has your commits in there too.  I want to do a few things on this since operating in this memory space is just hard.

I would say - if you want to take yours over the line, go for it.  I will rebase, if not, I'll do what I can.  Also, if you have the bandwidth to keep an eye on it for your own use cases I'd appreciate it.

https://github.com/Admiral-Piett/goaws/pull/282

Issue #270, opened 2023-09-07
Support SNS PublishBatch


perrydunn:
The equivalent SQS SendMessageBatch is supported.

I'm happy to implement, please let me know?

Admiral-Piett:
@perrydunn Good idea!  If you want to take on the implementation that would be awesome - feel free to assign this issue to yourself and make a PR?  If not, I'll get to it when I can.  :) 

For Reference: https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SendMessageBatch.html

perrydunn:
Sorry didn't mean to close! I don't appear to be able to assign this to myself but I am working on this now.

Admiral-Piett:
@perrydunn Excellent!  Thank you!

PR #274, opened 2023-11-03
Implement SNS:PublishBatch

perrydunn:
tests to be added
(already using with a project successfully)

### DIFF
--- a/app/servertest/server_test.go
+++ b/app/servertest/server_test.go
@@ -12,6 +12,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/Admiral-Piett/goaws/app"
+	"github.com/Admiral-Piett/goaws/app/router"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/session"
@@ -19,8 +21,6 @@ import (
 	"github.com/aws/aws-sdk-go/service/sqs"
 	"github.com/aws/aws-sdk-go/service/sqs/sqsiface"
 	"github.com/gorilla/mux"
-	"github.com/Admiral-Piett/goaws/app"
-	"github.com/Admiral-Piett/goaws/app/router"
 	log "github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -153,6 +153,24 @@ func TestSNSRoutes(t *testing.T) {
 	publishResponse, err := client.Publish(publishParams)
 	require.NoError(t, err, "SNS Publish Failed")
 	t.Logf("Succesfully published: %s\n", *publishResponse.MessageId)
+
+	publishBatchParams := &sns.PublishBatchInput{
+		TopicArn: response.TopicArn,
+		PublishBatchRequestEntries: []*sns.PublishBatchRequestEntry{
+			{
+				Id:      aws.String("1"),
+				Message: aws.String("Cool"),
+			},
+			{
+				Id:      aws.String("2"),
+				Message: aws.String("Dog"),
+			},
+		},
+	}
+	publishBatchResponse, err := client.PublishBatch(publishBatchParams)
+	require.NoError(t, err, "SNS PublishBatch Failed")
+	assert.Empty(t, publishBatchResponse.Failed)
+	assert.Length(t, publishBatchResponse.Successful, 2)
 }
 
 func newSQS(t *testing.T, region string, endpoint string) *sqs.SQS {


perrydunn:
ref https://github.com/Admiral-Piett/goaws/issues/270

app/gosns/gosns.go
```diff
@@ -485,6 +491,132 @@ func DeleteTopic(w http.ResponseWriter, req *http.Request) {
 
 }
 
+func PublishBatch(w http.ResponseWriter, req *http.Request) {
```
Admiral-Piett, line 494: This looks great - would you mind also updating the Readme on it?  Like adding a box and checking it [HERE](https://github.com/Admiral-Piett/goaws/blob/master/.github/README.md?plain=1#L43-L53), and updating the Postman collection with an example request [HERE](https://github.com/Admiral-Piett/goaws/blob/master/.github/README.md?plain=1#L103)?

app/gosns/gosns.go
```diff
@@ -485,6 +491,132 @@ func DeleteTopic(w http.ResponseWriter, req *http.Request) {
...
+		return
+	}
+
+	batchMessageIdToMessageBody := make(map[string]string, 10)
+	batchMessageIdToMessageStructure := make(map[string]string, 10)
+	batchMessageIdToMessageAttributes := make(map[string]map[string]app.MessageAttributeValue, 10)
+	batchMessageIdToSubject := make(map[string]string, 10)
+
+	for memberIndex := 1; true; memberIndex++ {
+		if memberIndex > 10 {
```
Admiral-Piett, line 512: So we have an implied limit of 10 on this?  Would it be worth doing a couple things for user clarity?
1. Return that in the errors?  `TooMany` is good, but would something like `Over10EntriesRequested` be more clear?  (Also, I'm not married to the camel case errors, especially in the responses - I know that's the current pattern  in here, but I leave you the option if you prefer.)
2. Is there a way we could bump this `Max 10` validation up before the loops and return it there?  If so, would that make it clearer when we loop by doing something like 
     `memberIndex := 1; memberIndex < len(members); memberIndex++`
If that works would that make it more clear what's happening in these loops?

app/gosns/gosns.go
```diff
@@ -485,6 +491,132 @@ func DeleteTopic(w http.ResponseWriter, req *http.Request) {
...
+			return
+		}
+
+		thisMessageBody := req.FormValue(thisMessageFormKey + ".Message")
+		thisMessageStructure := req.FormValue(thisMessageFormKey + ".MessageStructure")
+		thisMessageSubject := req.FormValue(thisMessageFormKey + ".Subject")
+
+		thisMessageAttributes := make(map[string]app.MessageAttributeValue)
+
+		for i := 1; true; i++ {
```
Admiral-Piett, line 535: Depending on your thoughts on the comment above, would it be clearer to throw a quick comment in here on what were trying to do?  I know we're setting the MessageAttributes, but it may take folks (me 😅) some digging to figure that out with the names here.

app/gosns/gosns.go
```diff
@@ -485,6 +491,132 @@ func DeleteTopic(w http.ResponseWriter, req *http.Request) {
...
+			}
+
+			dataType := req.FormValue(fmt.Sprintf("%s.MessageAttributes.entry.%d.Value.DataType", thisMessageFormKey, i))
+			if dataType == "" {
+				log.Warnf("DataType of %s.MessageAttribute %s is missing, MD5 checksum will most probably be wrong!\n", thisMessageFormKey, name)
+				continue
+			}
+
+			// StringListValue and BinaryListValue is currently not implemented
+			for _, valueKey := range [...]string{"StringValue", "BinaryValue"} {
```
Admiral-Piett, line 548: If someone provided both `StringValue` and `BinaryValue` would we overwrite `StringValue`?  If so, is that what we want?

Admiral-Piett (commented):
Looks great!  I had a couple minor questions, curious what you think about them when you have time.  Let me know about those tests and I'll get it marked for the next feature release!  

Thanks for taking this on!

app/gosns/gosns.go
perrydunn: It's an actual SNS limit - see [here](https://docs.aws.amazon.com/sns/latest/api/API_PublishBatch.html)
My approach has been to largely mimic the implemented SNS:Publish (gosns.Publish) and SQS:SendMessageBatch (gosqs...) for consistency.

app/gosns/gosns.go
perrydunn: How can I update the Postman collection?

app/gosns/gosns.go
perrydunn: Reading the docs this needs refinement. I took this from the SNS:Publish (gosns.Publish) implementation (which may also need some rework I'm not going to touch it here).

perrydunn: EDIT: Correction, the SNS docs are unclear on the behaviour here. I'll do what the localstack solution does...

app/gosns/gosns.go
Admiral-Piett: Are you using Postman to interact with goaws at all?  If you're not, I can update it for you, if you can give me a copy of your request with `curl` for reference.  It's located here - https://www.getpostman.com/collections/091386eae8c70588348e. Postman can import it if you just give it the link.

If you are, you can just include your new request in that collection and right click > select "Share" > Select "Via API" then replace the link here https://github.com/Admiral-Piett/goaws/blob/master/.github/README.md?plain=1#L113C65-L113C124 with your new link.

Happy to do it for you though if you prefer.

app/gosns/gosns.go
Admiral-Piett: Sounds good, I like your pattern here better anyway - it's clearer about what we're doing.  We can dig deeper on how exactly it's supposed to work later.

app/gosns/gosns.go
Admiral-Piett: Ah nice!  I see your entry in the error message now, so I think that serves.  Thanks!

app/gosns/gosns.go
Admiral-Piett: In fact - it might be better if you just send me the curl?  I'm going to have to a big update there looks like.  😮‍💨

Admiral-Piett:
Hm, I think we're close here - looks like the tests are failing on `assert.Length` which I don't see as a function in the testify library?  I think that's probably it.

Admiral-Piett:
@perrydunn quick follow up here.  Any thoughts on this?

PR #278, opened 2023-11-09
Fix erroneous "not found" responses in DeleteMessageBatch

adventureisyou:
When looping through messages, the "deleted" indicator was being updated in the copy of deleteEntry within the loop rather than updating the slice which is later used to append "not found" errors, so both successes and fails were being returned.

### DIFF
--- a/app/gosqs/gosqs.go
+++ b/app/gosqs/gosqs.go
@@ -644,16 +644,16 @@ func DeleteMessageBatch(w http.ResponseWriter, req *http.Request) {
 
 	app.SyncQueues.Lock()
 	if _, ok := app.SyncQueues.Queues[queueName]; ok {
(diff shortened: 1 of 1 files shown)


Admiral-Piett:
@adventureisyou Your change for this looks good, but I think we also need to include a fix for this line.  https://github.com/Admiral-Piett/goaws/pull/278/files#diff-e4a9df642f75a9a1287dc9408df6d0de9c9680711caa7d621db50bdd832e4daeR668.  I think it needs to be `if !deleteEntry.Deleted` - otherwise we'd still get the deleted entries in both places.  

Would you mind taking care of that update, and throwing a test or so here for it?

Admiral-Piett:
@adventureisyou Quick follow up on this, any thoughts on my previous comment?

PR #282, opened 2023-11-22
Refactor config.go

Admiral-Piett:
This also includes commits from @apavanello https://github.com/Admiral-Piett/goaws/pull/261.


Admiral-Piett:
I like this pattern but I'm going to have to rework the shared/in-memory resources before going further with fleshing this out.  Closing for now.

PR #284, opened 2023-11-28
Allow setting any log level

stanleyrh:
It would be useful to set a specific log level by providing a string as other levels apart from info and debug are useful.

Have prioritised the existing debug flag for backwards compatibility but would recommend its removal.

### DIFF
--- a/app/cmd/goaws.go
+++ b/app/cmd/goaws.go
@@ -18,6 +18,7 @@ import (
 func main() {
 	var filename string
 	var debug bool
+	var loglevel string
 	flag.StringVar(&filename, "config", "", "config file location + name")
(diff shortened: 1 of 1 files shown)


Admiral-Piett (approved):
Nice!

Admiral-Piett (commented):
This change looks good, but looks like the tests pointed out that we need to declare the variable first in case it's not passed in.  Mind taking care of that?

stanleyrh:
Ah my bad! Was editing straight in the browser and didn't run the tests 🤦 
Have fixed now

Admiral-Piett:
@stanleyrh nice!  Merging it!

PR #286, opened 2024-01-05
[WIP] Adding support for aws-json for SQS

mancej:
This is an active WIP, but I can confirm Send/Receive are at least working :) 

Please feel free to help me finish the transformation as the new AWS SDK versions only use the new JSON based SQS protocol so without adding support finished this library will require users to pin to very old versions of the SDK.

The goal here was to avoid refactors to any existing logic, but instead intercept the request or response and change the parsing logic to support both JSON and XML form based requests.


Admiral-Piett:
@mancej This looks awesome, thank you for working on this!  I CC'd you over here - https://github.com/Admiral-Piett/goaws/pull/285#pullrequestreview-1811381187 on the other PR around this that I think you noticed.  Some of the topics will be related I think.  I will try to keep the discussion over there for now, just for sanity.  😄  Thanks again for the support on this!

Admiral-Piett:
Working on this in an alternate feature branch

PR #287, opened 2024-01-07
Include Owner value in Subscriptions and SubscriptionsByTopic responses

chrisdaly3:
I added the `Owner` field for both the `ListSubscriptions` and `ListSubscriptionsByTopic` requests, as well as added tests to confirm the field is populated with the expected value in the response.

---
-> `ListSubscriptions`

image

---
-> `ListSubscriptionsByTopic`

image

> Note - sorry it took so long to get to this one!! Luckily the changes were pretty straightforward, and the tests are passing smoothly. Let me know if you have any questions 👍


Admiral-Piett:
This looks great to me, thanks for looking into it!  I rebased it and re-ran the checks, if those pass I shall merge it!

PR #288, opened 2024-01-12
[WIP] Feature aws json

Admiral-Piett:
WIP for the AWS json full feature.

**NOTE** 
If you plan to contribute to this feature, please PR against this branch.  We designed the first supported endpoint to (hopefully) provide a foundation/pattern on which to build the subsequent changes.  Once we get them all, we'll release this. 

To make sure we don't trip on each other, please update the list below by putting your name next to the API you intend to work on.  Once it gets into PR we can `x` that API and move on.  Thank you in advance for any support/tests/reviews/code work!

**JSON Supported APIs**
SQS
 - [x] CreateQueue - @Admiral-Piett 
 - [x] ListQueues - @Admiral-Piett 
 - [x] GetQueueAttributes - @Admiral-Piett 
     - (unsupported attributes are mocked)
 - [x] SetQueueAttributes - @Admiral-Piett 
     - (Only supported attributes are set - see Supported Queue Attributes)
 - [x] SendMessage - @kojisaikiAtSony 
 - [x] SendMessageBatch - @kojisaikiAtSony 
 - [x] ReceiveMessage - @andrew-womeldorf 
 - [x] DeleteMessage - @andrew-womeldorf 
 - [x] DeleteMessageBatch - @kojisaikiAtSony 
 - [x] GetQueueUrl - @kojisaikiAtSony 
 - [x] PurgeQueue - @Admiral-Piett 
 - [x] Delete Queue - @Admiral-Piett 
 - [x] ChangeMessageVisibility - @andrew-womeldorf 

SNS
 - [x] ListTopics - @kojisaikiAtSony 
 - [x] CreateTopic - @kojisaikiAtSony 
 - [x] DeleteTopic - @kojisaikiAtSony 
 - [x] Subscribe - @Admiral-Piett 
 - [x] SetSubscriptionAttributes - @kojisaikiAtSony 
 - [x] GetSubscriptionAttributes - @kojisaikiAtSony 
 - [x] ListSubscriptionsByTopic - @kojisaikiAtSony 
 - [x] ListSubscriptions - @kojisaikiAtSony 
 - [x] Unsubscribe - @Admiral-Piett 
 - [x] Publish - @Admiral-Piett 

SNS Internal
 - [x] ConfirmSubscription - @kojisaikiAtSony 

**Steps For PR:**
1. Create a model for your new Request struct
   a. Implement `<NewStruct>.SetAttributesFromForm` - even if it just returns without doing anything.
[...]

PR #289, opened 2024-01-17
Refactor for CreateQueue for V1 JSON support

Admiral-Piett:
This is a pretty rough draft of what I had in mind for our previous conversations.  If we like where this is headed, I'll fix it up a bit, rename everything away from just `V1`, and fix all the TODO's.  

Preliminarily I think it works for both XML and JSON flows.


Matthew-Davey:
| AWS SDK | goaws | Pass/Fail |
|---|---|---|
| 3.7.200.65 | admiralpiett/goaws:v0.4.5 | :heavy_check_mark: |
| 3.7.300.39 | admiralpiett/goaws:v0.4.5 | :heavy_multiplication_x: |
| 3.7.200.65 | feature-aws-json-v1 `docker build .` | :heavy_check_mark: |
| 3.7.300.39 | feature-aws-json-v1 `docker build .` | :heavy_multiplication_x: |

It looks like a lot of progress has been made here (:smiley:), I can see in the docker logs that many SQS operations seem to be successful, however my tests are failing at the teardown stage as they are unable to purge queues.

`docker logs goaws`

```
{"level":"info","msg":"Putting Message in Queue: callback-service-queue","time":"2024-01-22T11:57:34Z"}
{"level":"info","msg":"2024-01-22 11:57:34: Queue: callback-service-queue, Message: {...},"time":"2024-01-22T11:57:34Z"}
{"level":"info","msg":"Getting Message from Queue: callback-service-queue","time":"2024-01-22T11:57:34Z"}
{"level":"info","msg":"Deleting Message, Queue: callback-service-queue , ReceiptHandle: 3d777ccc-e9e8-4a89-b123-9020dd9b40d2#a1a55fe4-cd92-4635-9a40-f86e4a818a04","time":"2024-01-22T11:57:34Z"}
{"level":"info","msg":"FIFO Queue callback-service-queue unlocking group :","time":"2024-01-22T11:57:34Z"}
{"level":"info","msg":"Purging Queue: ","time":"2024-01-22T11:59:30Z"}
{"level":"info","msg":"Purge Queue:  , queue does not exist!!!","time":"2024-01-22T11:59:30Z"}
```

Admiral-Piett:
@Matthew-Davey Could you give me some info on how you are testing this?  Like - which SDK are you using?  Also, have you got this automated somehow?  (If so that would be huuuuuuuge here.  I've been meaning to do that for ages. 😄 )

I'm looking through the SDK's now, and the python one at least (`boto3`) still seems to be sending Forms for the time being, at least as far as I can tell.  I want to make sure I get on board with this contract before I take this much further.

Admiral-Piett:
@Matthew-Davey Following up on the above, some information would be helpful when you can.  Thanks!

Admiral-Piett:
Ok, I think [ffa95eb](https://github.com/Admiral-Piett/goaws/pull/289/commits/ffa95ebb59e34a696d9845640ce3c9b1674bb939) is the last of the required code for the CreateQueue action and the basis of the others.  Though I still want to smoke test this against an SDK that's doing JSON now.  All the one's I've tried say they support it but aren't actually making calls with it - at least not to CreateQueue.  Feedback/word on this would be appreciated.

NOTE: there are things we can improve on later and if you see things feel free to shout/make issues/etc., but I think this is workable for now.

In the meantime, future endpoints more or less could be accomplished with the following steps.  (Probably missing some steps, but this should be the bulk - I will update if we find things)
1. Create a model for your new Request struct
   a. Implement `<NewStruct>.SetAttributesFromForm` - even if it just returns without doing anything.
2. Create a `RequestFunctionV1` function
3. Swap the reference in the routing table to the new `V1` routing table
4. Copy the code from the old function
5. Call `utils.REQUEST_TRANSFORMER` with your new struct
6. Update fetches from the request Form to look at the new struct
7. Make the new V1 function return a status and response struct
8. Update any references to `createErrorResponse` to use `createErrorResponseV1`
9. Delete old/non-V1 function
10. Update/Create/Remove unit tests
11. Smoke Test with the SDK
12. PR

kojisaikiAtSony:
@dhumphreys01 @Admiral-Piett 
The smoke-test mechanism included in this PR is great!
However, there are many differences in one PR, and it is mixed with the discussion of aws-json, making it difficult to review.

How about creating a separate PR corresponding to smoke-test and adding it to `master` bransh first?

kojisaikiAtSony (commented):
@Admiral-Piett @dhumphreys01 
I think this idea would work! 👍 

My ideal architecture was to solve the protocol concerns around the router.go and have gosqs/gosns.go not depend on the http package (For example, the argument of CreateQueueV1 would be models.CreateQueueRequest).

However, my idea will be much bigger than this, so I think it is better to apply your current idea to other SQS/SNS APIs and test it working first!

kojisaikiAtSony:
I tried V1 flow on SQS.SendMessage API (There is a TODO on the attribute resolving but...) on #291.

I don't think the amount of code required to rewrite one API is too large 👍

kojisaikiAtSony:
https://github.com/Admiral-Piett/goaws/pull/289#issuecomment-1949372726
> 11. Smoke Test with the SDK
> 12. Add automated tests in smoke_tests/...

@Admiral-Piett 
I found https://github.com/Admiral-Piett/goaws/blob/master/app/servertest/server_test.go has end-to-end level testing with AWS SDK.
Is this an alternative to the smoke test you want to achieve?

A concern with testing with the AWS SDK is that we can only use one SDK version, so we cannot test `aws-json` and `aws-query` at the same time.

kojisaikiAtSony:
It may be a bit rough, but by taking advantage of the fact that there are v1 and v2 of separeted AWS SDK Go, we might be able to test aws-query using SDK v1 `1.47.6` and aws-json using SDK v2 `1.28.0`.
I'll try this idea separately.

kojisaikiAtSony:
> It may be a bit rough, but by taking advantage of the fact that there are v1 and v2 of separeted AWS SDK Go, we might be able to test aws-query using SDK v1 `1.47.6` and aws-json using SDK v2 `1.28.0`. I'll try this idea separately.

I created this PoC https://github.com/Admiral-Piett/goaws/pull/292

Admiral-Piett:
> > It may be a bit rough, but by taking advantage of the fact that there are v1 and v2 of separeted AWS SDK Go, we might be able to test aws-query using SDK v1 `1.47.6` and aws-json using SDK v2 `1.28.0`. I'll try this idea separately.
> 
> I created the PoC on the E2E testing idea: #292
> 
> In the PoC, I was convinced that it was necessary to separate the protocol context from `gosqs` and `gosns`...

@kojisaikiAtSony I really liked your ideas I used several of them, though I think my style differed slightly.  If you prefer some style changes that will make it easier to consume (or any others of course) just shout - I'm open to chatting about it.  I  I tested this against both styles of the SDK (luckily the python one hasn't updated to JSON and I had that handy, along with the Golang one), looks good prelimiarily.  The smoke tests are broken still - still trying to figure out why the servers are sharing the only memory addresses...sigh.  

Otherwise, I really liked your idea about incorporating the SDK into the automated tests - so I'm going to try to work on that at least for the JSON requests and go from there.  After that I think it's about there.  So let me know what you think about this?  If you agree, hopefully this will put us in a good space to knock out the rest more easily.

Thanks for sticking with this!

kojisaikiAtSony:
> I got the memory issue in the smoke tests too I think - watch out for the global memory resources. They are hold-over even across instances of the server. I wrote a utility to wipe them that you can call at the end of tests until I can figure out a long term strategy
Nice workaround! 👍 

> I really liked your idea about incorporating the SDK into the automated tests - so I'm going to try to work on that at least for the JSON requests and go from there
OK!

The latest end-to-end app behavior looks good. I also think the testing approach is good, so please continue the work 👍 
The only thing I'm concerned about is that the xml response from gosqs.go is further converted to json (the xml protocol infiltrates the SQS business layer). This is simply because I'm too particular about details. 😓 

Next time, I'll try updating the SendMessage implementation (https://github.com/Admiral-Piett/goaws/pull/291). At that time, I will consider the idea of layer division a little.

Admiral-Piett:
@kojisaiki Ok, this is ready for review I think - finally.  You may have noticed my other question out here - https://github.com/Admiral-Piett/goaws/pull/288#issuecomment-1979569818.  Let me know what you think there?  I don't want to step on your toes, but these changes look fairly old now.  

If we can remove them then this is ready for your opinion and hopefully we can get it merged shortly and divide and conquer the rest of it.

kojisaikiAtSony:
> @kojisaiki Ok, this is ready for review I think - finally. You may have noticed my other question out here - [#288 (comment)](https://github.com/Admiral-Piett/goaws/pull/288#issuecomment-1979569818). Let me know what you think there? I don't want to step on your toes, but these changes look fairly old now.
> 
> If we can remove them then this is ready for your opinion and hopefully we can get it merged shortly and divide and conquer the rest of it.
No problem, you can remove my old commits/codes completely that are from https://github.com/Admiral-Piett/goaws/pull/285 (I closed the PR) 👍

smoke_tests/sqs_create_queue_test.go
```diff
@@ -0,0 +1,447 @@
...
+		utils.ResetResources()
+	}()
+
+	utils.InitializeDecoders()
+
+	e := httpexpect.Default(t, server.URL)
+
+	e.POST("/").
+		WithHeaders(map[string]string{
+			"Content-Type": "application/x-amz-json-1.0",
```
kojisaikiAtSony, line 374: The test name is `...xml...`, but the test request looks using `aws-json`.

kojisaikiAtSony (changes requested):

smoke_tests/sqs_create_queue_test.go
kojisaikiAtSony: It seems that the test name and the protocol actually used for the other tests are different, so could you please review them?

Admiral-Piett: That is just to create a basic queue first so that I can set it as the Dead Letter Queue in the attributes below.  If that queue doesn't exist when I make the second request it will error.  But it's not particularly important in and of itself.  

Below, the next queue is created with XML which is the one with attributes populated - and the one I am asserting on at the end.

PR #293, opened 2024-03-14
[aws-json] Migrate SQS SendMessage

kojisaikiAtSony:
- [x] Create a model for your new Request struct
    - a. Implement <NewStruct>.SetAttributesFromForm - even if it just returns without doing anything.
    - b. You may also need to handle UnmarshalJSON for any nested request attributes/objects.
- [x] Create a RequestFunctionV1 function
- [x] Swap the reference in the routing table to the new V1 routing table
- [x] Copy the code from the old function
- [x] Call utils.REQUEST_TRANSFORMER with your new struct
- [x] Update fetches from the request Form to look at the new struct
- [x] Make the new V1 function return a status and response struct
- [x] Update any references to createErrorResponse to use createErrorResponseV1
- [x] Delete old/non-V1 function
- [x] Update/Create/Remove unit tests
- [x] Smoke Test with the SDK
- [x] Add automated tests in smoke_tests/... (use the v2 SDK as much as possible)

### DIFF
--- a/app/gosqs/gosqs.go
+++ b/app/gosqs/gosqs.go
@@ -42,7 +42,7 @@ func init() {
 	app.SqsErrors["InvalidVisibilityTimeout"] = err8
 	err9 := app.SqsErrorType{HttpError: http.StatusBadRequest, Type: "MessageNotInFlight", Code: "AWS.SimpleQueueService.MessageNotInFlight", Message: "The message referred to isn't in flight."}
 	app.SqsErrors["MessageNotInFlight"] = err9
-	err10 := app.SqsErrorType{HttpError: http.StatusBadRequest, Type: "MessageTooBig", Code: "InvalidMessageContents", Message: "The message size exceeds the limit."}
+	err10 := app.SqsErrorType{HttpError: http.StatusBadRequest, Type: "MessageTooBig", Code: "InvalidParameterValue", Message: "The message size exceeds the limit."}
 	app.SqsErrors["MessageTooBig"] = err10
 	app.SqsErrors[ErrInvalidParameterValue.Type] = *ErrInvalidParameterValue
 	app.SqsErrors[ErrInvalidAttributeValue.Type] = *ErrInvalidAttributeValue
@@ -92,92 +92,6 @@ func PeriodicTasks(d time.Duration, quit <-chan struct{}) {
 			ticker.Stop()
 			return
 		}
-	}
-}
-
-func SendMessage(w http.ResponseWriter, req *http.Request) {
-	w.Header().Set("Content-Type", "application/xml")
-	req.ParseForm()
-	messageBody := req.FormValue("MessageBody")
-	messageGroupID := req.FormValue("MessageGroupId")
-	messageDeduplicationID := req.FormValue("MessageDeduplicationId")
-	messageAttributes := extractMessageAttributes(req, "")
-
-	queueUrl := getQueueFromPath(req.FormValue("QueueUrl"), req.URL.String())
-
-	queueName := ""
-	if queueUrl == "" {
-		vars := mux.Vars(req)
-		queueName = vars["queueName"]
-	} else {
-		uriSegments := strings.Split(queueUrl, "/")
-		queueName = uriSegments[len(uriSegments)-1]
-	}
-
-	if _, ok := app.SyncQueues.Queues[queueName]; !ok {
-		// Queue does not exist
-		createErrorResponse(w, req, "QueueNotFound")
-		return
-	}
-
-	if app.SyncQueues.Queues[queueName].MaximumMessageSize > 0 &&
-		len(messageBody) > app.SyncQueues.Queues[queueName].MaximumMessageSize {
-		// Message size is too big
-		createErrorResponse(w, req, "MessageTooBig")
-		return
-	}
-
-	delaySecs := app.SyncQueues.Queues[queueName].DelaySeconds
-	if mv := req.FormValue("DelaySeconds"); mv != "" {
-		delaySecs, _ = strconv.Atoi(mv)
-	}
-
-	log.Println("Putting Message in Queue:", queueName)
-	msg := app.Message{MessageBody: []byte(messageBody)}
-	if len(messageAttributes) > 0 {
-		msg.MessageAttributes = messageAttributes
-		msg.MD5OfMessageAttributes = common.HashAttributes(messageAttributes)
-	}
-	msg.MD5OfMessageBody = common.GetMD5Hash(messageBody)
-	msg.Uuid, _ = common.NewUUID()
-	msg.GroupID = messageGroupID
-	msg.DeduplicationID = messageDeduplicationID
-	msg.SentTime = time.Now()
-	msg.DelaySecs = delaySecs
-
-	app.SyncQueues.Lock()
-	fifoSeqNumber := ""
-	if app.SyncQueues.Queues[queueName].IsFIFO {
-		fifoSeqNumber = app.SyncQueues.Queues[queueName].NextSequenceNumber(messageGroupID)
-	}
-
-	if !app.SyncQueues.Queues[queueName].IsDuplicate(messageDeduplicationID) {
-		app.SyncQueues.Queues[queueName].Messages = append(app.SyncQueues.Queues[queueName].Messages, msg)
-	} else {
-		log.Debugf("Message with deduplicationId [%s] in queue [%s] is duplicate ", messageDeduplicationID, queueName)
-	}
-
-	app.SyncQueues.Queues[queueName].InitDuplicatation(messageDeduplicationID)
-	app.SyncQueues.Unlock()
-	log.Infof("%s: Queue: %s, Message: %s\n", time.Now().Format("2006-01-02 15:04:05"), queueName, msg.MessageBody)
-
-	respStruct := app.SendMessageResponse{
-		Xmlns: "http://queue.amazonaws.com/doc/2012-11-05/",
-		Result: app.SendMessageResult{
-			MD5OfMessageAttributes: msg.MD5OfMessageAttributes,
-			MD5OfMessageBody:       msg.MD5OfMessageBody,
-			MessageId:              msg.Uuid,
-			SequenceNumber:         fifoSeqNumber,
-		},
-		Metadata: app.ResponseMetadata{
-			RequestId: "00000000-0000-0000-0000-000000000000",
-		},
-	}
-
-	enc := xml.NewEncoder(w)
-	enc.Indent("  ", "    ")
-	if err := enc.Encode(respStruct); err != nil {
-		log.Printf("error: %v\n", err)
 	}
 }
 
--- a/app/gosqs/send_message.go
+++ b/app/gosqs/send_message.go
@@ -1,0 +1,133 @@
+package gosqs
+
+import (
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/Admiral-Piett/goaws/app/interfaces"
+	"github.com/Admiral-Piett/goaws/app/models"
+
+	"github.com/Admiral-Piett/goaws/app/utils"
+
+	log "github.com/sirupsen/logrus"
+
+	"github.com/Admiral-Piett/goaws/app"
+	"github.com/Admiral-Piett/goaws/app/common"
+	"github.com/gorilla/mux"
+)
+
+func SendMessageV1(req *http.Request) (int, interfaces.AbstractResponseBody) {
+	requestBody := models.NewSendMessageRequest()
+	ok := utils.REQUEST_TRANSFORMER(requestBody, req)
+	if !ok {
+		log.Error("Invalid Request - CreateQueueV1")
+		return createErrorResponseV1(ErrInvalidParameterValue.Type)
+	}
+	messageBody := requestBody.MessageBody
+	messageGroupID := requestBody.MessageGroupId
+	messageDeduplicationID := requestBody.MessageDeduplicationId
+	messageAttributes := requestBody.MessageAttributes
+
+	queueUrl := getQueueFromPath(requestBody.QueueUrl, req.URL.String())
+
+	queueName := ""
+	if queueUrl == "" {
+		// TODO: Remove this query param logic if it's not still valid or something
+		vars := mux.Vars(req)
+		queueName = vars["queueName"]
+	} else {
+		uriSegments := strings.Split(queueUrl, "/")
+		queueName = uriSegments[len(uriSegments)-1]
+	}
+
+	if _, ok := app.SyncQueues.Queues[queueName]; !ok {
+		// Queue does not exist
+		return createErrorResponseV1("QueueNotFound")
+	}
+
+	if app.SyncQueues.Queues[queueName].MaximumMessageSize > 0 &&
+		len(messageBody) > app.SyncQueues.Queues[queueName].MaximumMessageSize {
+		// Message size is too big
+		return createErrorResponseV1("MessageTooBig")
+	}
+
+	delaySecs := app.SyncQueues.Queues[queueName].DelaySeconds
+	if requestBody.DelaySeconds != 0 {
+		delaySecs = requestBody.DelaySeconds
+	}
+
+	log.Debugf("Putting Message in Queue: [%s]", queueName)
+	msg := app.Message{MessageBody: []byte(messageBody)}
+	if len(messageAttributes) > 0 {
+		oldStyleMessageAttributes := convertToOldMessageAttributeValueStructure(messageAttributes)
+		msg.MessageAttributes = oldStyleMessageAttributes
+		msg.MD5OfMessageAttributes = common.HashAttributes(oldStyleMessageAttributes)
+	}
+	msg.MD5OfMessageBody = common.GetMD5Hash(messageBody)
+	msg.Uuid, _ = common.NewUUID()
+	msg.GroupID = messageGroupID
+	msg.DeduplicationID = messageDeduplicationID
+	msg.SentTime = time.Now()
+	msg.DelaySecs = delaySecs
+
+	app.SyncQueues.Lock()
+	fifoSeqNumber := ""
+	if app.SyncQueues.Queues[queueName].IsFIFO {
+		fifoSeqNumber = app.SyncQueues.Queues[queueName].NextSequenceNumber(messageGroupID)
+	}
+
+	if !app.SyncQueues.Queues[queueName].IsDuplicate(messageDeduplicationID) {
+		app.SyncQueues.Queues[queueName].Messages = append(app.SyncQueues.Queues[queueName].Messages, msg)
+	} else {
+		log.Debugf("Message with deduplicationId [%s] in queue [%s] is duplicate ", messageDeduplicationID, queueName)
+	}
+
+	app.SyncQueues.Queues[queueName].InitDuplicatation(messageDeduplicationID)
+	app.SyncQueues.Unlock()
+	log.Infof("%s: Queue: %s, Message: %s\n", time.Now().Format("2006-01-02 15:04:05"), queueName, msg.MessageBody)
+
+	respStruct := models.SendMessageResponse{
+		Xmlns: "http://queue.amazonaws.com/doc/2012-11-05/",
+		Result: models.SendMessageResult{
+			MD5OfMessageAttributes: msg.MD5OfMessageAttributes,
+			MD5OfMessageBody:       msg.MD5OfMessageBody,
+			MessageId:              msg.Uuid,
+			SequenceNumber:         fifoSeqNumber,
+		},
+		Metadata: app.ResponseMetadata{
+			RequestId: "00000000-0000-0000-0000-000000000000",
+		},
+	}
+
+	return http.StatusOK, respStruct
+}
+
+// TODO:
+// Refactor internal model for MessageAttribute between SendMessage and ReceiveMessage
+// from app.MessageAttributeValue(old) to models.MessageAttributeValue(new) and remove this temporary function.
+func convertToOldMessageAttributeValueStructure(newValues map[string]models.MessageAttributeValue) map[string]app.MessageAttributeValue {
+	attributes := make(map[string]app.MessageAttributeValue)
+
+	for name, entry := range newValues {
+		// StringListValue and BinaryListValue is currently not implemented
+		// Please refer app/gosqs/message_attributes.go
+		value := ""
+		valueKey := ""
+		if entry.StringValue != "" {
+			value = entry.StringValue
+			valueKey = "StringValue"
+		} else if entry.BinaryValue != "" {
+			value = entry.BinaryValue
+			valueKey = "BinaryValue"
+		}
+		attributes[name] = app.MessageAttributeValue{
+			Name:     name,
+			DataType: entry.DataType,
+			Value:    value,
+			ValueKey: valueKey,
+		}
+	}
+
+	return attributes
+}
--- a/app/gosqs/send_message_test.go
+++ b/app/gosqs/send_message_test.go
@@ -1,0 +1,219 @@
+package gosqs
+
+import (
+	"net/http"
+	"testing"
+	"time"
+
+	"github.com/Admiral-Piett/goaws/app"
+	"github.com/Admiral-Piett/goaws/app/fixtures"
+	"github.com/Admiral-Piett/goaws/app/interfaces"
+	"github.com/Admiral-Piett/goaws/app/models"
+	"github.com/Admiral-Piett/goaws/app/utils"
+	"github.com/stretchr/testify/assert"
+)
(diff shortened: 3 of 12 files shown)


app/gosqs/send_message.go
kojisaikiAtSony, line 1: I'm trying to separate files on each API since `gosqs.go` looks very long file. It may cause a conflict in migrating each API in parallel.

I wanted to create `send_message_test.go` extracted from `gosqs_test.go`. However, `gosqs_test.go` contains some "scenario like" test. So, now I keep the existing test in original file.

app/models/send_message_model.go
kojisaikiAtSony, line 1: Why don't we create separated model files that contains Request/Response on each API?
It is good to avoid the conflict and easy to find the code point on each API.

kojisaikiAtSony:
Sorry to late but I'm still alive 👍 . I want to update this within 2 or 3 weeks...

app/models/send_message_model.go
Admiral-Piett: For each model?  To me, that seems a little granular.  I might leave the models together for now and see how we feel.  I liked your idea above for the handlers though I so I started breaking them up in the `ListQueues` PR.  See what you think?

app/gosqs/send_message.go
Admiral-Piett: Good thinking, I added this kind of thing to the `ListQueues` PR, seems cleaner.  Good idea!

app/models/send_message_model.go
kojisaikiAtSony: Yep, `ListQueue` looks the diff on model files are not so much. OK, I will update my PR as like `ListQueue` 👍

app/gosqs/gosqs.go
```diff
@@ -42,7 +42,7 @@ func init() {
 	app.SqsErrors["InvalidVisibilityTimeout"] = err8
 	err9 := app.SqsErrorType{HttpError: http.StatusBadRequest, Type: "MessageNotInFlight", Code: "AWS.SimpleQueueService.MessageNotInFlight", Message: "The message referred to isn't in flight."}
 	app.SqsErrors["MessageNotInFlight"] = err9
-	err10 := app.SqsErrorType{HttpError: http.StatusBadRequest, Type: "MessageTooBig", Code: "InvalidMessageContents", Message: "The message size exceeds the limit."}
+	err10 := app.SqsErrorType{HttpError: http.StatusBadRequest, Type: "MessageTooBig", Code: "InvalidParameterValue", Message: "The message size exceeds the limit."}
```
kojisaikiAtSony, line 45: In today's actual SQS behavior and AWS SDK, the code was changed to InvalidParameterValue.

app/gosqs/send_message.go
```diff
@@ -0,0 +1,133 @@
...
+		},
+	}
+
+	return http.StatusOK, respStruct
+}
+
+// TODO:
+// Refactor internal MessageAttribute model between SendMessage and ReceiveMessage
+// from app.MessageAttributeValue(old) to models.MessageAttributeValue(new) and remove this temporary function.
+func convertToOldMessageAttributeValueStructure(newValues map[string]models.MessageAttributeValue) map[string]app.MessageAttributeValue {
```
kojisaikiAtSony, line 109: Current internal MessageAttribute model `app.MessageAttributeValue` is a bit difference with recent [official MessageAtteributeValue](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_MessageAttributeValue.html).

If there is any matter to use the current struct after migration all the SQS/SNS API, we can refactor the `app.MessageAttributeValue` to latest `models.MessageAttributeValue`.

kojisaikiAtSony:
Hi @Admiral-Piett, this is ready for review 👍

app/gosqs/send_message_test.go
```diff
@@ -0,0 +1,111 @@
...
+		&app.Queue{Name: "test_max_message_size", MaximumMessageSize: 100}
+
+	form := url.Values{}
+	form.Add("Action", "SendMessage")
+	form.Add("QueueUrl", "http://localhost:4100/queue/test_max_message_size")
+	form.Add("MessageBody", "test%20message%20body%201")
+	form.Add("Version", "2012-11-05")
+	req.PostForm = form
+
+	status, response := SendMessageV1(req)
```
Admiral-Piett, line 29: Checkout how I did these kinds of tests.  Let's get into that pattern if we can.  I intentionally made `utils. REQUEST_TRANSFORMER` a global variable so we could swap it in tests at run time.  Just so every unit test didn't pass through that each time.  

I guess it's fine if they do, but this request handling stuff, I want to get it out of the handler logic as much as we can - still chewing on how to do it ultimately, but for now, I prefer this pattern below?

https://github.com/Admiral-Piett/goaws/blob/feature-aws-json/app/gosqs/create_queue_test.go#L30-L34

app/gosqs/send_message_test.go
```diff
@@ -0,0 +1,111 @@
...
+	if status != http.StatusOK {
+		t.Errorf("handler returned wrong status code: got %v want %v",
+			status, http.StatusOK)
+	}
+
+	// Check the response body is what we expect.
+	sendMessageResponse, ok := response.(models.SendMessageResponse)
+	assert.True(t, ok)
+	if len(sendMessageResponse.Result.MD5OfMessageBody) == 0 {
+		t.Errorf("handler returned unexpected body: got %v",
```
Admiral-Piett, line 41: Let's use `assert` and check the entire message body, at least just once.

app/gosqs/send_message_test.go
```diff
@@ -0,0 +1,111 @@
...
+	"net/http"
+	"net/url"
+	"testing"
+
+	"github.com/Admiral-Piett/goaws/app"
+	"github.com/Admiral-Piett/goaws/app/models"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestSendMessage_MaximumMessageSize_Success(t *testing.T) {
```
Admiral-Piett, line 13: Would it be clearer to just call it `TestSendMessage_Success`?  `MaximumMessageSize` doesn't seem to factor in much?

app/gosqs/send_message_test.go
```diff
@@ -0,0 +1,111 @@
...
+	req, err := http.NewRequest("POST", "/", nil)
+	if err != nil {
+		t.Fatal(err)
```
Admiral-Piett, line 16: `assert.Nil`

app/gosqs/send_message_test.go
Admiral-Piett: Also, getting a good `FIFO` sequence increment?  At least in one of these?

app/gosqs/send_message_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+	// Check the response body is what we expect.
+	errorResponse, ok := response.(models.ErrorResponse)
+	assert.True(t, ok)
+	expected := "Not Found"
+	if errorResponse.Result.Type != expected {
+		t.Errorf("handler returned unexpected body: got %v want %v",
+			errorResponse.Result.Type, expected)
+	}
+}
```
Admiral-Piett, line 111: Can we also do a test on what happens if the request transformer fails?

https://github.com/Admiral-Piett/goaws/blob/feature-aws-json/app/gosqs/create_queue_test.go#L239-L254

app/models/models.go
```diff
@@ -153,6 +153,62 @@ func (r *ListQueueRequest) SetAttributesFromForm(values url.Values) {
...
+}
+type MessageAttributeValue struct {
+	BinaryListValues []string `json:"BinaryListValues"` // goaws does not supported yet
+	BinaryValue      string   `json:"BinaryValue"`
+	DataType         string   `json:"DataType"`
+	StringListValues []string `json:"StringListValues"` // goaws does not supported yet
+	StringValue      string   `json:"StringValue"`
+}
+
+func (r *SendMessageRequest) SetAttributesFromForm(values url.Values) {
```
Admiral-Piett, line 182: Looks like this could use a test?  Something like below?

https://github.com/Admiral-Piett/goaws/blob/feature-aws-json/app/models/models_test.go#L39-L85

smoke_tests/fixtures/requests.go
```diff
@@ -7,6 +7,16 @@ import (
...
 
+var CreateQueuesRequestBodyXML = struct {
+	Action    string `xml:"Action"`
+	Version   string `xml:"Version"`
+	QueueName string `xml:"QueueName"`
+}{
+	Action:    "CreateQueue",
+	Version:   "2012-11-05",
+	QueueName: "new-queue-1",
+}
```
Admiral-Piett, line 18: Do we still need this?  I thought I took this out.  I think it's ok to use the JSON sdk for creating queues when testing other things?  I only particularly care about making sure the XML works in the dedicated smoke tests.

Admiral-Piett (commented):
Great job!  I had a few questions around, but lets talk about them if you have questions.  I'm trying to get a few patterns going here, but if you have better ideas I'm all ears.

Thanks for all your hard work!

app/gosqs/send_message.go
```diff
@@ -0,0 +1,133 @@
...
+	messageGroupID := requestBody.MessageGroupId
+	messageDeduplicationID := requestBody.MessageDeduplicationId
+	messageAttributes := requestBody.MessageAttributes
+
+	queueUrl := getQueueFromPath(requestBody.QueueUrl, req.URL.String())
+
+	queueName := ""
+	if queueUrl == "" {
+		vars := mux.Vars(req)
+		queueName = vars["queueName"]
```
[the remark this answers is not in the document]
kojisaikiAtSony: Sorry, I don't know because it was an existing code.
I'm not sure how Go's `http.Request`'s `Context` is set, but all existing tests succeeded even if I removed this mux part.

It may be possible to investigate separately from aws-json support.

app/gosqs/send_message.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	if app.SyncQueues.Queues[queueName].MaximumMessageSize > 0 &&
+		len(messageBody) > app.SyncQueues.Queues[queueName].MaximumMessageSize {
+		// Message size is too big
+		return createErrorResponseV1("MessageTooBig")
+	}
+
+	delaySecs := app.SyncQueues.Queues[queueName].DelaySeconds
+	if mv := req.FormValue("DelaySeconds"); mv != "" {
```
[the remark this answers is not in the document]
kojisaikiAtSony: Oh yes, we already have DelaySeconds in `models.SendMessageRequest`. I'll use it 👍

app/gosqs/send_message.go
```diff
@@ -0,0 +1,133 @@
...
+		delaySecs, _ = strconv.Atoi(mv)
+	}
+
+	log.Println("Putting Message in Queue:", queueName)
```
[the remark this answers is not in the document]
kojisaikiAtSony: It was existing code, but, agree 👍

app/models/models.go
```diff
@@ -153,6 +153,62 @@ func (r *ListQueueRequest) SetAttributesFromForm(values url.Values) {
...
+func NewSendMessageRequest() *SendMessageRequest {
+	return &SendMessageRequest{
+		MessageAttributes:       make(map[string]MessageAttributeValue),
+		MessageSystemAttributes: make(map[string]MessageAttributeValue),
+	}
+}
+
+type SendMessageRequest struct {
+	DelaySeconds            int                              `json:"Del1aySeconds" schema:"DelaySeconds"`
+	MessageAttributes       map[string]MessageAttributeValue `json:"MessageAttributes" schema:"MessageAttributes"`
```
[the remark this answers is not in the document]
kojisaikiAtSony: Yes, `MessageAttributes` and `MessageSystemAttributes` are the ability to append any fields on users demand.
And sorry, I haven't checked the names for `MessageSystemAttributes` is defined by AWS. 

How do you think to mark `MessageSystemAttributes` as "unsupported" with definition of current `map[string]MessageAttributeValue` at this time? 
Because AFAIK this "goaws" is not support emulation of AWS X-Ray. 
We can also add support later if we or users find a MessageSystemAttribute name between SQS/SNS.

app/gosqs/send_message_test.go
kojisaikiAtSony: This test clearly sets `MaximumMessageSize` with a large value to succeed the sending.
image
But I think there is no problem to rename 👍

app/gosqs/send_message_test.go
kojisaikiAtSony: OK!
To be honest, I don't usually write Go at work, so I think there are some aspects of the culture that I don't understand.
For testing, `assert.Nil` is often used rather than `t.Fatal`, right?

smoke_tests/fixtures/requests.go
kojisaikiAtSony: I think it's okay to delete it. At least in the SendMessage test, CreateQueue using the SDK works fine, including setting MaximumMessageSize!
I believe that testing with XML has been confirmed in existing tests even before we started this work.

app/gosqs/send_message_test.go
kojisaikiAtSony: (Now in my TODO)

app/models/models.go
kojisaikiAtSony: (Now in my TODO)

smoke_tests/fixtures/requests.go
kojisaikiAtSony: I realized `sqs_create_queue_test.go` is the only test that use XML (form protocol) since `create_queue_test.go` only has UT that is separated from any protocol.

If future goaws also supports XML, I think we should keep it for XML testing. However, for other tests such as SendMessage, it is sufficient to be able to create a queue with the SDK, so I think this XML definition can be moved into `sqs_create_queue_test.go` as private (starting with a lowercase letter).

app/models/models.go
kojisaikiAtSony: Added!

kojisaikiAtSony:
@Admiral-Piett I updated the PR!

smoke_tests/fixtures/requests.go
kojisaikiAtSony: Finally, I haven't used it. I removed it 👍

app/gosqs/send_message_test.go
kojisaikiAtSony: I updated the UTs!
https://github.com/Admiral-Piett/goaws/pull/293#discussion_r1601627025

app/sqs_messages.go
```diff
@@ -15,6 +14,17 @@ type SendMessageResponse struct {
...
+/*** List Queues Response */
+type ListQueuesResult struct {
+	QueueUrl []string `xml:"QueueUrl"`
+}
+
+type ListQueuesResponse struct {
+	Xmlns    string           `xml:"xmlns,attr"`
+	Result   ListQueuesResult `xml:"ListQueuesResult"`
+	Metadata ResponseMetadata `xml:"ResponseMetadata"`
+}
```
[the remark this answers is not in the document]
kojisaikiAtSony: Correct, we have already moved `app.---` to `models.---` completely. I removed from here.

app/models/models.go
```diff
@@ -153,6 +153,62 @@ func (r *ListQueueRequest) SetAttributesFromForm(values url.Values) {
...
+
+func NewSendMessageRequest() *SendMessageRequest {
+	return &SendMessageRequest{
+		MessageAttributes:       make(map[string]MessageAttributeValue),
+		MessageSystemAttributes: make(map[string]MessageAttributeValue),
+	}
+}
+
+type SendMessageRequest struct {
+	DelaySeconds            int                              `json:"Del1aySeconds" schema:"DelaySeconds"`
```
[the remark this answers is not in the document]
kojisaikiAtSony: I added `DelaySceconds` in the requests in smoke tests.
image
This conversion seems to be handled well by XmlDecoder.

smoke_tests/sqs_send_message_test.go
```diff
@@ -0,0 +1,220 @@
...
+	sdkConfig, _ := config.LoadDefaultConfig(context.TODO())
+	sdkConfig.BaseEndpoint = aws.String(server.URL)
+	sqsClient := sqs.NewFromConfig(sdkConfig)
+	sdkResponse, _ := sqsClient.CreateQueue(context.TODO(), &sqs.CreateQueueInput{
+		QueueName: &af.QueueName,
+	})
+	targetQueueUrl := sdkResponse.QueueUrl
+
+	// Assert no messages in the queue before sending message
+	getQueueAttributesBodyXML := struct {
```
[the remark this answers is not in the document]
kojisaikiAtSony: Refactored! 👍

smoke_tests/sqs_send_message_test.go
```diff
@@ -0,0 +1,220 @@
...
+		Action     string `xml:"Action"`
+		Version    string `xml:"Version"`
+		Attribute1 string `xml:"AttributeName.1"`
```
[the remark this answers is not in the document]
kojisaikiAtSony: Refactored with SDK 👍

smoke_tests/sqs_send_message_test.go
```diff
@@ -0,0 +1,220 @@
...
+		}
+	}
+
+	// Target test: send a message
+	targetMessageBody := "Test_SendMessageV1_json_no_attributes"
+	sendMessageOutput, _ := sqsClient.SendMessage(context.TODO(), &sqs.SendMessageInput{
+		QueueUrl:    targetQueueUrl,
+		MessageBody: &targetMessageBody,
+	})
+	assert.NotNil(t, sendMessageOutput.MessageId)
```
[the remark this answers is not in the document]
kojisaikiAtSony: Added!

smoke_tests/sqs_send_message_test.go
```diff
@@ -0,0 +1,220 @@
...
+	r = e.POST("/").
+		WithForm(receiveMessageBodyXML).
+		Expect().
+		Status(http.StatusOK).
+		Body().Raw()
+	r3 := app.ReceiveMessageResponse{}
+	xml.Unmarshal([]byte(r), &r3)
+	message := r3.Result.Message[0]
+	assert.Equal(t, targetMessageBody, string(message.Body))
+	assert.Equal(t, 2, len(message.MessageAttributes))
```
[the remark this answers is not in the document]
kojisaikiAtSony: Added assert with String, Number and Binary! 👍

app/gosqs/send_message_test.go
kojisaikiAtSony: Updated!

app/gosqs/send_message_test.go
```diff
@@ -0,0 +1,219 @@
...
+	// Check the response
+	assert.Equal(t, http.StatusOK, status)
+	sendMessageResponse, ok := response.(models.SendMessageResponse)
+	assert.True(t, ok)
+	assert.NotEmpty(t, sendMessageResponse.Result.MD5OfMessageBody)
+	// Should have FIFO Sequence
+	assert.NotEmpty(t, sendMessageResponse.Result.SequenceNumber)
+}
+
+func TestSendMessageV1_Success_Deduplication(t *testing.T) {
```
kojisaikiAtSony, line 97: I have added a test for Message Deduplication, but it looks goaws is not supported the latest AWS interface, right?

On current code:
Apply `EnableDuplicates` on the Queue

Latest AWS:
Apply `DeduplicationScope` on the Queue
https://docs.aws.amazon.com/ja_jp/AWSSimpleQueueService/latest/APIReference/API_CreateQueue.html#API_CreateQueue_RequestParameters

app/gosqs/send_message_test.go
Admiral-Piett: Haha - thank you!  Sorry if it was confusing.  Trying get the non-handler specific logic abstracted away was tricky.

app/gosqs/send_message_test.go
```diff
@@ -0,0 +1,219 @@
...
+	status, response := SendMessageV1(r)
+
+	// Check the status code is what we expect.
+	assert.Equal(t, http.StatusBadRequest, status)
+
+	// Check the response body is what we expect.
+	errorResponse, ok := response.(models.ErrorResponse)
+	assert.True(t, ok)
+	assert.Equal(t, "Not Found", errorResponse.Result.Type)
+}
```
Admiral-Piett, line 219: These look great!  Optional: Do we want to do maybe just 1 more test with some attributes?

app/models/models.go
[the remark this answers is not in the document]
Admiral-Piett: Ok sounds good - if you pushed it through the SDKs that's cool with me.  Like I said, different SDKs did different things for me, so just something to keep an eye on.

app/models/models_test.go
```diff
@@ -255,6 +255,42 @@ func TestListQueuesRequest_SetAttributesFromForm_invalid_max_results(t *testing.
...
+	form.Add("MessageAttribute.1.Value.StringValue", "Value1")
+	form.Add("MessageAttribute.2.Name", "Attr2")
+	form.Add("MessageAttribute.2.Value.DataType", "String")
+	form.Add("MessageAttribute.2.Value.BinaryValue", "VmFsdWUy")
+	form.Add("MessageAttribute.3.Name", "")
+	form.Add("MessageAttribute.3.Value.DataType", "String")
+	form.Add("MessageAttribute.3.Value.StringValue", "Value")
+	form.Add("MessageAttribute.4.Name", "Attr4")
+	form.Add("MessageAttribute.4.Value.DataType", "")
+	form.Add("MessageAttribute.4.Value.StringValue", "Value4")
```
Admiral-Piett, line 271: Would it make sense to make at least one of these fields a `BinaryValue`?

app/sqs_messages.go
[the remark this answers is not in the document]
Admiral-Piett: Awesome, thank you!  I know it's a lot of juggling.

Admiral-Piett (commented):
This is awesome!  Thank you so much for keeping up with all my comments.  This looks great.  I had a handful of tiny comments just curious about your response on a few of them.  Otherwise I think this is pretty much ready, no?

Admiral-Piett:
Oh, one other tiny thing, would you mind squashing your commits?  I think it's easier in the long run to track things when there's 1 commit per "change".

app/gosqs/send_message_test.go
kojisaikiAtSony: Sorry, what test would you like to add? For MessageAttributes? Or is it any of Queue Attribute?

app/models/models_test.go
kojisaikiAtSony: The test for "Attr2" is that 👍

app/gosqs/send_message.go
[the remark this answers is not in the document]
Admiral-Piett: Awesome, thank you!  I see now it's all over the place...I'll have to think about this more.  But I really appreciate the support!

app/gosqs/send_message_test.go
Admiral-Piett: No worries!  I appreciate your flexibility here, especially on my nitpicks! 😄   

I like `assert.<whatever>` since that's a benefit of the `testify` library.  It makes the error states more uniform and easier to decipher, for me anyway.  `t.Fatal` / `t.Error` is Go's native testing pattern, but I think it's more clunky to deal with and lot trickier to figure out which statement is the issue if there are several in a single test.  It also requires a lot of hardcoded error codes, etc., which I don't think are valuable.

app/gosqs/send_message_test.go
Admiral-Piett: Yeah good point - probably outdated.  I'll try to get to that at some point.  I only asked for the test because the code supports the skipping of the message if it's a duplicate, but you're right - in real life, it could use some work.

Admiral-Piett (approved):
This looks great!  The only outstanding thing I see is the MessageSystemAttributes thing we talked about here - https://github.com/Admiral-Piett/goaws/pull/293#discussion_r1607225650  

If you want to just hold on that that's probably fine.  It's probably a headache to refactor and I can live with it following the `MessageAttributes` pattern for now, especially since it's not functional.   

Otherwise, it looks terrific!  Thank you for all your hard work here!  Let me know what you think about the above and I'll kick off the tests and get it merged?

app/models/models.go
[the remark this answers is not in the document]
kojisaikiAtSony: Sorry to miss the response 🙇 
I think we don't have to change `map[string]` for `MessageSystemAttributes` soon since the latest AWS SDK accept the `map[string]` for it.
image

I'll mark it as TODO.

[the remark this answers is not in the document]
Admiral-Piett: Yeah you can leave it for now.  I think it would be easier to manage long term with a nested struct, but I can deal with it later.  Thanks anyway!

PR #294, opened 2024-03-15
Fix FIFO group locking behaviour

### LINKED ISSUE #212: SQS batch not working.
I cannot receive batch massages always getting single message at time. Sample request
Action=ReceiveMessage&Version=2012-11-05&AttributeName.1=All&MessageAttributeName.1=All&MaxNumberOfMessages=10&WaitTimeSeconds=20.

I debugged the code. gosqs.go 

if app.SyncQueues.Queues[queueName].IsFIFO {
				// If we got messages here it means we have not processed it yet, so get next
				if app.SyncQueues.Queues[queueName].IsLocked(msg.GroupID) {
					continue
				}
				// Otherwise lock messages for group ID
				app.SyncQueues.Queues[queueName].LockGroup(msg.GroupID)
			}
This runs in loop and only the first message will returned. The second iteration will execute.
if app.SyncQueues.Queues[queueName].IsLocked(msg.GroupID) {
					continue
				}

Thomasvdam:
First of all thank you for this amazing software, it's already been incredibly helpful! While using it I noticed that my FIFO batches were always only a single message and after some investigation I noticed it was due to a small bug in the implementation.

I added a (perhaps too verbose) test to verify that my suspicion was correct and that the fix actually worked but that uncovered a bunch of other bugs in the FIFO group locking behaviour so the PR became a little bigger than intended.

This is my first serious Golang contribution so my apologies if it's not idiomatic or performant Go.

## Explanation of changes

Previously a FIFO queue was only able to return 1 message per request as it locked the group right after encountering the first message. Group unlocking had a similar problem where any deleted message in a group would unlock the whole group, even if there were pending messages in that group.

In addition messages could receive a receipt/timeout despite not being served if they were part of a locked group. By moving the receipt assignment to come after the group lock check this should no longer occur.

## Related issues

I think this should close #212


Admiral-Piett:
@Thomasvdam Thanks for taking a look at this.  Overall the goal of the change looks good to me, but we're in the middle of a large scale feature.  I can't let this in since it'll cause more deployment headaches than we can bare there right now.  I'm sorry.  Once we're finished with the JSON work, let's regroup on this and see how this fits in with whatever comes out of that.

Thanks again for supporting us here - I really appreciate it.

Thomasvdam:
No problem at all! I’ll keep an eye out for the JSON update 😃

PR #308, opened 2024-06-25
Add SNS SubscribeV1 for new pattern support

Admiral-Piett:


app/utils/utils.go
```diff
@@ -72,3 +74,18 @@ func ExtractQueueAttributes(u url.Values) map[string]string {
 	}
 	return attr
 }
+
+func CreateErrorResponseV1(errKey string, isSqs bool) (int, interfaces.AbstractResponseBody) {
```
kojisaikiAtSony, line 78: I like to use type alias like `AwsProtocol`.
e.g.
```
type ErrorType int

const (
	SqsError  ErrorType = iota
	SnsError ErrorType = iota
)
```

app/models/models.go
kojisaikiAtSony, line 1: Why don't we move`AVAILABLE_QUEUE_ATTRIBUTES` also to `models/sqs.go` ?

kojisaikiAtSony (approved):

kojisaikiAtSony:
@Admiral-Piett There are a few of comments but totally looks good! I will start to modify my CreateTopic PR to merge with this branch 👍

app/conf/mock-data/mock-config.yaml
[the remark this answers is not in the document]
Admiral-Piett: Just because they're all already on the default one, `/conf/goaws.yaml`, and I got tired of having them all over the place.  I figured one source of reference would be enough?

app/examples/java/SnsSample.java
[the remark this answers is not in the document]
Admiral-Piett: Yeah - I think they were a throw to old automated smoke tests.  I think we have our new pattern now which will work.  If we want to incorporate more SDKs down the road we can - I think we can try to cross that bridge when we need it.

app/utils/utils.go
Admiral-Piett: Let's hold there - there's only 2 options at the moment.  If we need more than that, or it gets too confusing, we can swap to another pattern like the `AwsProtocol`.  How does that sound?

app/models/models.go
Admiral-Piett: Well, we could.  I just wasn't sure if it would be shared eventually or parts of it, and it's more of a constant than a model - I just couldn't think of anywhere else to put it.  If, at the end, it's not shared then I'll move it?

app/conf/mock-data/mock-config.yaml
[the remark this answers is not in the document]
kojisaikiAtSony: It makes sense! 👍

app/utils/utils.go
kojisaikiAtSony: It's OK!

app/models/models.go
kojisaikiAtSony: OK, we don't have to move it soon 👍

PR #309, opened 2024-07-03
[aws-json] Migrate SQS SendMessageBatch

kojisaikiAtSony:


app/gosqs/send_message_batch.go
```diff
@@ -0,0 +1,108 @@
...
+		}
+		ids[v.Id] = struct{}{}
+	}
+
+	sentEntries := make([]models.SendMessageBatchResultEntry, 0)
+	log.Debug("Putting Message in Queue:", queueName)
+	for _, sendEntry := range sendEntries {
+		msg := app.Message{MessageBody: []byte(sendEntry.MessageBody)}
+		if len(sendEntry.MessageAttributes) > 0 {
+			oldStyleMessageAttributes := convertToOldMessageAttributeValueStructure(sendEntry.MessageAttributes)
```
kojisaikiAtSony, line 65: I'm using the conversion step as SendMessageV1
https://github.com/Admiral-Piett/goaws/pull/288/files#diff-2b6e3902e73e67c2bf76a6c2ee3f3cc4be55d19105a543cf7954186e99c1ef54R63

app/models/sqs.go
```diff
@@ -222,6 +223,71 @@ func (r *SendMessageRequest) SetAttributesFromForm(values url.Values) {
...
+	for key := range values {
+
+		keySegments := strings.Split(key, ".")
+		if len(keySegments) <= 3 {
+			// no attribute
+			continue
+		}
+		entryIndex, err1 := strconv.Atoi(keySegments[1])
+		attributeIndex, err2 := strconv.Atoi(keySegments[3])
+		if err1 != nil || err2 != nil {
```
kojisaikiAtSony, line 245: The case the number of key segments will be 4 (index 0 to 3) like: `Entries.1.MessageAttributes.1.Name`
The case that the number will be under 3 like: `Entries.1.MessageBody`

kojisaikiAtSony:
Hi @Admiral-Piett, could you please review this?

app/models/sqs.go
Admiral-Piett: Ah I see.  Could you make a note of why the 2 are there?  There's no way I'm going to remember that the Message Attributes have extra layers to them.

Admiral-Piett (approved):
Awesome!  1 tiny request for a comment in the code and all set.  Thanks a ton for all your hard work here!  These are heroic!

kojisaikiAtSony:
Ready to merge!

app/models/sqs.go
```diff
@@ -222,6 +223,74 @@ func (r *SendMessageRequest) SetAttributesFromForm(values url.Values) {
...
+	for key := range values {
+
+		keySegments := strings.Split(key, ".")
+		//If keysegments is 3 or less, there is no attribute value like: (Entries.1.MessageBody)
+		if len(keySegments) <= 3 {
+			continue
+		}
+
+		// From like a value(Entries.1.MessageAttributes.1.Name), get the entry index and the attribute index.
+		entryIndex, err1 := strconv.Atoi(keySegments[1])
```
Admiral-Piett, line 245: I'm sorry, can we have a comment that mentions the pattern you're parsing at this index, and then another one for the index below?  This will be difficult to consume and remember in the long term and the documentation is poor at describing it.  

It took me a while in searching to figure out what we were going for (no fault of yours!), and I'd like to avoid that 30 minute search down the road.

Admiral-Piett: Even just something like the below would be amazing.  Thank you!
```
Both patterns below are supported here.
// strconv.Atoi(keySegments[1] - targets the index value in pattern: `Entries.1.MessageBody`
// strconv.Atoi(keySegments[3] - targets the index value in pattern: `Entries.1.MessageAttributes.1.Name`
```

kojisaikiAtSony:
@Admiral-Piett Updated!

PR #310, opened 2024-07-09
[aws-json] Migrate SQS DeleteMessageBatch

kojisaikiAtSony:


app/gosqs/delete_message_batch.go
```diff
@@ -0,0 +1,114 @@
...
+
+	if _, ok := app.SyncQueues.Queues[queueName]; !ok {
+		return utils.CreateErrorResponseV1("QueueNotFound", true)
+	}
+
+	if len(requestBody.Entries) == 0 {
+		return utils.CreateErrorResponseV1("EmptyBatchRequest", true)
+	}
+
+	if len(requestBody.Entries) > 10 {
```
Admiral-Piett, line 44: Where did you find this requirement?  I don't see it in the docs or in the previous code, but I could be missing it.

app/gosqs/delete_message_batch.go
```diff
@@ -0,0 +1,114 @@
...
+			Deleted:       false,
+		}
+	}
+
+	app.SyncQueues.Lock()
+	defer app.SyncQueues.Unlock()
+
+	deletedEntries := make([]models.DeleteMessageBatchResultEntry, 0)
+	for j, deleteEntry := range deleteEntries {
+		for i, msg := range app.SyncQueues.Queues[queueName].Messages {
```
Admiral-Piett, line 70: I know this was the existing code, but this pattern feels labor intensive.  If nothing else I think it might be more performant to 1. fetch out the queue once, and 2. at least flip the order of these loops if not use a data structure that doesn't need the 2 loops, like a map.  If we just flipped the order, we'd know that we only have a max of 10 entries (unless we remove that requirement above), in `deleteEntries`.  In the messages, in theory, we could have an unlimited amount.

You don't have to change it now, but would you mind putting `# TODO - Admiral-Piett` with a link to this comment in there, so I can come back to it at some point?

app/gosqs/delete_message_batch.go
```diff
@@ -0,0 +1,114 @@
...
+			if msg.ReceiptHandle == deleteEntry.ReceiptHandle {
+				// Unlock messages for the group
+				log.Printf("FIFO Queue %s unlocking group %s:", queueName, msg.GroupID)
```
Admiral-Piett, line 73: Debug?

Admiral-Piett (commented):
Beautiful!  2 tiny questions as always, but otherwise this is great!  Much improved even!

kojisaikiAtSony:
> I think this one needs a commit squash too btw.

Yes, we are working on your comments, please wait a moment 👍

app/gosqs/delete_message_batch.go
kojisaikiAtSony: This is for "TooManyEntriesInBatchRequest" of the below:
https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_DeleteMessageBatch.html

And the API abstraction says "Deletes up to ten messages from the specified queue." 👍

app/gosqs/delete_message_batch.go
kojisaikiAtSony: @Admiral-Piett
Our friend, Mr. Otsuka(you can see the name on the commit 😁 ) did the refactoring!

- Create a Map called deleteMessageMap in advance, with ReceiptHandle as the key and deleteEntry as the value.
- Scan the messages in the Queue and match them with the ReceiptHandle of the deleteEntry to determine which messages to delete.
- If the ReceiptHandle matches, determine that the message is to be deleted and perform the delete process as before.

Admiral-Piett: Awesome!  Thank you both then!  That looks great!  I really appreciate it!

app/gosqs/delete_message_batch.go
Admiral-Piett: Cool, I did not read that first line I guess haha.  Thank you for being thorough!

Admiral-Piett:
Awesome work to you and your team!  Really appreciate it as always!  Let's do this one next if that's ok, and then we can rebase others and get them in too?  

I'm closing in on Publish and I see you guys took the rest, so we're about there.  Heroic effort on this project - happy to see the light at the end of the tunnel!

Admiral-Piett (approved):

PR #312, opened 2024-07-09
[aws-json] Migrate SNS DeleteTopic

kojisaikiAtSony:


app/gosns/delete_topic.go
```diff
@@ -0,0 +1,47 @@
...
+	if !ok {
+		log.Error("Invalid Request - DeleteTopicV1")
+		return utils.CreateErrorResponseV1("InvalidParameterValue", false)
+	}
+
+	topicArn := requestBody.TopicArn
+	uriSegments := strings.Split(topicArn, ":")
+	topicName := uriSegments[len(uriSegments)-1]
+
+	log.Println("Delete Topic - TopicName:", topicName)
```
Admiral-Piett, line 28: Can we make these all info/warning/debug/etc., whatever seems appropriate where we see them?  the `Print*` stuff doesn't abide by the set levels and is difficult to manage.

This one seems like `info` to me, but I don't feel strongly.

app/models/sns.go
```diff
@@ -66,3 +66,17 @@ type SubscriptionAttributes struct {
...
+func NewDeleteTopicRequest() *DeleteTopicRequest {
+	return &DeleteTopicRequest{}
+}
+
+type DeleteTopicRequest struct {
+	TopicArn string `json:"TopicArn" schema:"TopicArn"`
+}
+
+func (r *DeleteTopicRequest) SetAttributesFromForm(values url.Values) {
+	r.TopicArn = values.Get("TopicArn")
```
Admiral-Piett, line 81: I don't think this will hurt anything, but I think the `schema` tag will help it find this.  This method pattern is really for the stuff like `Attribute.1.name` > `Attribute.1.value` so we can find each thing and use the 2 parts across both pieces to allocate it properly.

So I might remove this unless you had another need for it?  It should be ok for the interface if the method is just empty.

Admiral-Piett:
Also squashing and rebasing?

Is there an order you want me to try to merge these in?  I expect you're going to have some conflicts as we go, but I will try to accommodate?

kojisaikiAtSony:
@Admiral-Piett 
Thank you for reviewing very concurrently 4 PRs!
When multiple PRs become squash/merge ready, can Icombine multiple PRs into one commit, one new PR?
Conflicts may occur every time you merge them one by one.

kojisaikiAtSony:
Ready to merge!

Admiral-Piett (approved):

PR #316, opened 2024-07-29
[aws-json] Fix the invalid format "subscriptionArn" value in Subscribe API response

kojisaikiAtSony:
# Problem
The actual response value "subscriptionArn" is not a valid arn format.

image

# Background
This bug will affects to the developments on other APIs that use Subscription.

### DIFF
--- a/smoke_tests/sns_subscribe_test.go
+++ b/smoke_tests/sns_subscribe_test.go
@@ -2,6 +2,7 @@ package smoke_tests
 
 import (
 	"context"
+	"encoding/xml"
 	"fmt"
 	"net/http"
 	"testing"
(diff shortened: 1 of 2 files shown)


smoke_tests/sns_subscribe_test.go
```diff
@@ -58,6 +60,55 @@ func Test_Subscribe_json(t *testing.T) {
 	assert.Equal(t, "sqs", subscriptions[0].Protocol)
 	assert.False(t, subscriptions[0].Raw)
 	assert.Contains(t, subscriptions[0].SubscriptionArn, fmt.Sprintf("%s:%s", af.BASE_SNS_ARN, "unit-topic2"))
+	assert.Equal(t, response.SubscriptionArn, &subscriptions[0].SubscriptionArn)
+}
+
+func Test_Subscribe_json_with_duplicate_subscription(t *testing.T) {
```
kojisaikiAtSony, line 66: This test asserts that the subscription should not be duplicated when the multiple same subscriptions are requested.

dhumphreys01 (approved):

Admiral-Piett (approved):

PR #321, opened 2024-09-03
[aws-json] Migrate SNS SetSubscriptionAttributes API

kojisaikiAtSony:


app/gosns/set_subscription_attributes.go
```diff
@@ -0,0 +1,74 @@
...
+		if err != nil {
+			return utils.CreateErrorResponseV1("InvalidParameterValue", false)
+		}
+		app.SyncTopics.Lock()
+		sub.FilterPolicy = filterPolicy
+		app.SyncTopics.Unlock()
+
+	default:
+		// FIXME: This response is literally strange but it's the same behavior with old SetSubscriptionAtteributes.
+		return utils.CreateErrorResponseV1("SubscriptionNotFound", false)
```
kojisaikiAtSony, line 54: Current behavior returns "SubscriptionNotFound" error when an unsupported attribute name is specified.
Should we define another appropriate error?

app/gosns/set_subscription_attributes.go
```diff
@@ -0,0 +1,74 @@
...
+func getSubscription(subsArn string) *app.Subscription {
+	for _, topic := range app.SyncTopics.Topics {
+		for _, sub := range topic.Subscriptions {
+			if sub.SubscriptionArn == subsArn {
+				return sub
+			}
+		}
+	}
+	return nil
+}
```
kojisaikiAtSony, line 74: I tried to separate the loop indent from the following logic.

app/gosns/set_subscription_attributes.go
```diff
@@ -0,0 +1,74 @@
...
+
+	default:
+		// FIXME: This response is literally strange but it's the same behavior with old SetSubscriptionAtteributes.
+		return utils.CreateErrorResponseV1("SubscriptionNotFound", false)
+	}
+
+	uuid := uuid.NewString()
+	respStruct := models.SetSubscriptionAttributesResponse{
+		Xmlns:    models.BASE_XMLNS,
+		Metadata: app.ResponseMetadata{RequestId: uuid}}
```
kojisaikiAtSony, line 60: I realized that the "RequestId" in the response is different between APIs (random uuid or  `models.BASE_RESPONSE_METADATA`).
We can align it into which one before merging the feature branch into master.

kojisaikiAtSony:
Hi @Admiral-Piett , this is ready for review!

app/gosns/set_subscription_attributes.go
Admiral-Piett: Yeah, I know.  They're all different between SQS and SNS.  Let's match to whatever is currently here right now, and I'll go back and untangle them later.

app/gosns/set_subscription_attributes.go
Admiral-Piett: Nice!  If you want to throw it in the utils or somewhere more general it might be helpful in the other handlers?

Admiral-Piett (commented):
Looks great to me, and great call outs on your comments!  I think if we clear our questions up this is ready.

app/gosns/set_subscription_attributes.go
kojisaikiAtSony: Agree! I will apply it 👍

app/gosns/set_subscription_attributes.go
kojisaikiAtSony: OK, I will apply the idea in this PR. At lease, I will apply it on GetSubscriptionAttributes API.

app/gosns/set_subscription_attributes.go
kojisaikiAtSony: Applied!

app/gosns/set_subscription_attributes.go
kojisaikiAtSony: I shared the `getSubscription` to just only GetSubscriptionAttributes API. 
Unsubscribe API logic also has the same loop and indent but the API requires the subscription index in a topic and modify them by topic level perspective. I thought it will be too much commonize for the API.

kojisaikiAtSony:
@Admiral-Piett Updated!

Admiral-Piett (approved):
Lovely!  Thank you for your diligence, great work my friend!

PR #322, opened 2024-09-05
[aws-json] Migrate SNS ConfirmSubscription API

kojisaikiAtSony:


app/gosns/confirm_subscription.go
```diff
@@ -0,0 +1,40 @@
...
+
+	if pendingConfirm.token != confirmToken {
+		return utils.CreateErrorResponseV1("SubscriptionNotFound", false)
+	}
+	respStruct := models.ConfirmSubscriptionResponse{
+		Xmlns:    models.BASE_XMLNS,
+		Result:   models.ConfirmSubscriptionResult{SubscriptionArn: pendingConfirm.subArn},
+		Metadata: app.ResponseMetadata{RequestId: uuid.NewString()},
+	}
+	return http.StatusOK, respStruct
```
kojisaikiAtSony, line 39: As far as we looked into current impl, current ConfirmSubscription do nothing actually since the subscriptions in goaws won't be "pending" status even if the subscribe endpoint will be an email address or a HTTP url.

kojisaikiAtSony:
Now we don't have any smoke test for ConfirmSubscription because to perform the complete ConfirmSubscription feature requires the email destination and HTTP server.
Additionally, currently Subscribe API does not support "pending" status and [don't send an email even if the endpoint is an email address](https://github.com/Admiral-Piett/goaws/blob/feature-aws-json/app/gosns/subscribe.go#L63-L92).

kojisaikiAtSony:
Hi @Admiral-Piett, we tried to migrate the API but we have few comments. Could you please check them?

app/gosns/confirm_subscription.go
Admiral-Piett: Yeah I think this endpoint is there as a functionality nicety.  I haven't looked at this in AWS too closely but it must double check on the HTTP subscriptions.  While it's not "doing" anything, you're right, it is just kind of checking giving you a check that we have a record of the subscription.  I guess that's worthy? 🤷

Admiral-Piett (approved):

Admiral-Piett:
> Now we don't have any smoke test for ConfirmSubscription because to perform the complete ConfirmSubscription feature requires the email destination and HTTP server. Additionally, currently Subscribe API does not support "pending" status and [don't send an email even if the endpoint is an email address](https://github.com/Admiral-Piett/goaws/blob/feature-aws-json/app/gosns/subscribe.go#L63-L92).

That's ok since this is an internal endpoint anyway.  I'm fine to skip that one for now.  Thank you!

app/gosns/confirm_subscription.go
kojisaikiAtSony: Oh, sorry I didn't want to say to remove this API. 
I think the existence of this API on goaws is valuable in itself. Without it, ConfirmSubscription requests from user's applications would return 404 and it would not be able to fulfill its role as a stub.
Please get my above comment was for to be on the same page about current implementation 👍

Admiral-Piett: Yeah, I think we're saying the same thing in different ways.  I think it's fine that there's no real pending status, we'll just do what we do here.  The word "pending" in the code, I think is also still fine here, since it shadows what would probably be case in real life.  So all good, as far as I'm concerned!

app/router/router.go
```diff
@@ -89,11 +89,9 @@ var routingTableV1 = map[string]func(r *http.Request) (int, interfaces.AbstractR
 	"GetSubscriptionAttributes": sns.GetSubscriptionAttributesV1,
 	"SetSubscriptionAttributes": sns.SetSubscriptionAttributesV1,
 	"ListSubscriptionsByTopic":  sns.ListSubscriptionsByTopicV1,
-}
 
-var routingTable = map[string]http.HandlerFunc{
```
kojisaikiAtSony, line 94: The old routing table will be retired!

Issue #328, opened 2024-10-22
GoAWS sends empty MessageAttributes & MD5OfMessageAttributes when none are supplied, breaking compatibiltiy with aws/aws-sdk-php


zikeji:
Relevant issue: https://github.com/aws/aws-sdk-php/issues/2827

When creating a message with no attributes, the message returned by receive message has an empty `MessageAttributes` object and a empty string for `MD5OfMessageAttributes`.

This can be  reproduced with:

```bash
aws sqs send-message --queue-url http://goaws:4100/100010001000/default --message-body "test"
# {
#    "MD5OfMessageBody": "098f6bcd4621d373cade4e832627b4f6",
#    "MD5OfMessageAttributes": "",
#    "MessageId": "d9e27ccb-bbd7-4569-b00f-a6f6d91bdb37",
#    "SequenceNumber": ""
# }
aws sqs receive-message --queue-url http://goaws:4100/100010001000/default
# {
#    "Messages": [
#        {
#            "MessageId": "d9e27ccb-bbd7-4569-b00f-a6f6d91bdb37",
#            "ReceiptHandle": "d9e27ccb-bbd7-4569-b00f-a6f6d91bdb37#94880fdc-9092-4ef4-bbf6d950b2ddc966",
#            "MD5OfBody": "098f6bcd4621d373cade4e832627b4f6",
#            "Body": "test",
#            "Attributes": {
#                "ApproximateFirstReceiveTimestamp": "1729637818936",
#                "ApproximateReceiveCount": "1",
#                "SenderId": "100010001000",
#                "SentTimestamp": "1729637818936"
#            },
#            "MD5OfMessageAttributes": "",
#            "MessageAttributes": {}
#        }
#    ]
# }
```

kojisaikiAtSony:
It is related with jackson-ize process with empty map in go 🤔 We may need to add a condition somewhere to prevent map from being made carelessly.

Admiral-Piett:
@zikeji I maaaay have gotten this with the 0.5.1 release.  Would you mind checking when you have time and letting me know what you find?  Thank you!

Admiral-Piett:
@zikeji Got it.  I thought that might happen, but wanted to make sure I was on the right track.  I think this should take care of it - https://github.com/Admiral-Piett/goaws/pull/335.  I'll get it in v0.5.2!

Admiral-Piett:
[...]

PR #330, opened 2024-10-24
[Bugfix] Add base64 decode on json mashaller

kojisaikiAtSony:
We found an issue with the binary handling of MessageAttributes.

It causes the md5 hash validation error with AWS SDK for Java. Surprisingly, AWS SDK for Go does not have the md5 hash validation...

### DIFF
--- a/app/models/responses.go
+++ b/app/models/responses.go
@@ -1,6 +1,7 @@
 package models
 
 import (
+	"encoding/base64"
 	"encoding/json"
 
 	"github.com/Admiral-Piett/goaws/app"
@@ -80,10 +81,12 @@ func (r *ResultMessage) MarshalJSON() ([]byte, error) {
 	}
 
 	for _, attr := range r.MessageAttributes {
+		// When encoding json, BinaryValue([]byte) are automatically converted to base64.
+		binary, _ := base64.StdEncoding.DecodeString(attr.Value.BinaryValue)
 		m.MessageAttributes[attr.Name] = sqstypes.MessageAttributeValue{
 			DataType:    &attr.Value.DataType,
-			StringValue: &attr.Value.StringValue,
-			BinaryValue: []byte(attr.Value.BinaryValue),
+			StringValue: attr.Value.StringValue,
+			BinaryValue: binary,
 		}
 	}
 
@@ -91,9 +94,9 @@ func (r *ResultMessage) MarshalJSON() ([]byte, error) {
 }
 
 type ResultMessageAttributeValue struct {
-	DataType    string `xml:"DataType,omitempty"`
-	StringValue string `xml:"StringValue,omitempty"`
-	BinaryValue string `xml:"BinaryValue,omitempty"`
+	DataType    string  `xml:"DataType,omitempty"`
+	StringValue *string `xml:"StringValue,omitempty"`
+	BinaryValue string  `xml:"BinaryValue,omitempty"`
 }
 
 type ResultMessageAttribute struct {
--- a/smoke_tests/sns_publish_test.go
+++ b/smoke_tests/sns_publish_test.go
@@ -2,7 +2,10 @@ package smoke_tests
 
 import (
 	"context"
-	"encoding/json"
+	"crypto/md5"
+	"encoding/binary"
+	"encoding/hex"
+	"hash"
 	"io"
 	"net/http"
 	"net/http/httptest"
@@ -24,8 +27,6 @@ import (
 	"github.com/Admiral-Piett/goaws/app"
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/sns"
-	"github.com/aws/aws-sdk-go-v2/service/sns/types"
-	"github.com/aws/aws-sdk-go-v2/service/sqs"
 	"github.com/stretchr/testify/assert"
 )
 
@@ -73,6 +74,12 @@ func Test_Publish_Sqs_With_Message_Attributes(t *testing.T) {
 		server.Close()
 		test.ResetResources()
 	}()
+	addBytesToHash := func(hasher hash.Hash, arr []byte) {
+		bs := make([]byte, 4)
+		binary.BigEndian.PutUint32(bs, uint32(len(arr)))
+		hasher.Write(bs)
+		hasher.Write(arr)
+	}
 
 	sdkConfig, _ := config.LoadDefaultConfig(context.TODO())
 	sdkConfig.BaseEndpoint = aws.String(server.URL)
@@ -89,18 +96,28 @@ func Test_Publish_Sqs_With_Message_Attributes(t *testing.T) {
 		Name: topicName,
 	})
 
-	snsClient.Subscribe(context.TODO(), &sns.SubscribeInput{
+	subscribeResult, _ := snsClient.Subscribe(context.TODO(), &sns.SubscribeInput{
 		Protocol:              aws.String("sqs"),
 		TopicArn:              createTopicResult.TopicArn,
 		Attributes:            map[string]string{},
 		Endpoint:              createQueueResult.QueueUrl,
 		ReturnSubscriptionArn: true,
 	})
+
+	snsClient.SetSubscriptionAttributes(context.TODO(), &sns.SetSubscriptionAttributesInput{
+		SubscriptionArn: subscribeResult.SubscriptionArn,
+		AttributeName:   aws.String("RawMessageDelivery"),
+		AttributeValue:  aws.String("true"),
+	})
+
 	message := "{\"IAm\": \"aMessage\"}"
+	binaryValueString := "binary-value"
+	binaryValue := []byte(binaryValueString)
 	subject := "I am a subject"
+
 	attributes := map[string]types.MessageAttributeValue{
 		"someKey": {
-			BinaryValue: []byte(message),
+			BinaryValue: binaryValue,
 			DataType:    aws.String("Binary"),
 		},
 	}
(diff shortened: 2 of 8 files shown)


app/models/responses.go
```diff
@@ -80,20 +81,22 @@ func (r *ResultMessage) MarshalJSON() ([]byte, error) {
...
 
 	for _, attr := range r.MessageAttributes {
+		// When encoding json, BinaryValue([]byte) are automatically converted to base64.
+		binary, _ := base64.StdEncoding.DecodeString(attr.Value.BinaryValue)
 		m.MessageAttributes[attr.Name] = sqstypes.MessageAttributeValue{
 			DataType:    &attr.Value.DataType,
-			StringValue: &attr.Value.StringValue,
-			BinaryValue: []byte(attr.Value.BinaryValue),
+			StringValue: attr.Value.StringValue,
+			BinaryValue: binary,
```
kojisaikiAtSony, line 89: Here is the key change of this PR.

`encoding/json` lib automatically applies base 64 encode if the arg type is `slice`.
But our internal `MessageAttributeValue.BinaryValue` is already base 64 encoded string.
So the base64 encode was done "twice".

smoke_tests/sns_publish_test.go
```diff
@@ -102,26 +121,33 @@ func Test_Publish_Sqs_With_Message_Attributes(t *testing.T) {
...
+
+	// check client md5hash and received md5hash
+	hasher := md5.New()
+	addBytesToHash(hasher, []byte("someKey"))
+	addBytesToHash(hasher, []byte(*messageAttribute.DataType))
+	hasher.Write([]byte{2})
+	addBytesToHash(hasher, messageAttribute.BinaryValue)
+	clientHash := hex.EncodeToString(hasher.Sum(nil))
+
+	assert.Equal(t, clientHash, *receiveMessageResponse.Messages[0].MD5OfMessageAttributes)
```
kojisaikiAtSony, line 149: Added md5 hash assertion.

smoke_tests/sqs_send_message_batch_test.go
```diff
@@ -318,27 +320,27 @@ func TestSendMessageBatchV1_Json_Success(t *testing.T) {
...
 				DataType:    *attr.DataType,
-				StringValue: *attr.StringValue,
+				StringValue: attr.StringValue,
 			}
 		}
 	}
 	assert.Equal(t, binaryAttribute, attr1.Name)
 	assert.Equal(t, binaryType, attr1.Value.DataType)
-	assert.Equal(t, "YmFzZTY0LWVuY29kZWQtdmFsdWU=", attr1.Value.BinaryValue) // base64 encoded value
+	assert.Equal(t, binaryValueString, attr1.Value.BinaryValue)
```
kojisaikiAtSony, line 335: Through the AWS SDK we should receive the raw string instead of the base 64 encoded string.

kojisaikiAtSony:
Hi @Admiral-Piett we found another issue. Could you please check this?

By the way, we found that this fix and the previous fix on nil check are all that is needed to be compatible with the latest AWS SDK for Java!

app/models/responses.go
Admiral-Piett: I need some more information about what's going on here.  I don't really mind not encoding this, but I've gotten several complaints about the MessageAttributes in the last week or so and I need to be more careful about it.  

Can you explain more about what the issue is that you found, and how this will solve it?  Also the below if you know:

- When did the `[]byte` get encoded to begin with?  If we're doing it internally, let's just avoid doing it at all, and let the marshallers handle it.

- It's been a minute since I looked at this, but why did we have to do the override of `MarshalJSON` period here?  I see your comment about the structure, but I can't see why we needed an array to begin with.  It seems to me like it comes in as a map in `SendMessage` why not maintain the map format and not kill ourselves going back and forth?  Was this to be compatible with legacy code some how?  I know there's all this "convert to old style" stuff, but I sense that's the essence of a lot of these problems.

Admiral-Piett (commented):
Hey @kojisaikiAtSony - thanks for looking at this but I need some more info.  See my response to your comment below.

Also, what's all this extra stuff if all we needed here was to make sure we're not double encoding something.  Thanks!

app/models/responses.go
kojisaikiAtSony: The error we faced was like below: 
```
com.amazonaws.AmazonClientException: MD5 returned by SQS does not match the calculation on the original request. (MD5 calculated by the message attributes: "565a1514fba6a202b2f3e2de433300cf", MD5 checksum returned: "1dffa6b307b25550e81d62cf3377dacd")
```

> When did the []byte get encoded to begin with?
The base 64 encoding was performed by AWS SDK Client, and goaws preserves the base 64 encoded string received over HTTP for internal processing. So the value was already encoded at here.

> It's been a minute since I looked at this, but why did we have to do the override of MarshalJSON period here?
It's unclear whether we can completely get rid of MarshalJSON while supporting both xml and json, but as you say, tackling the below "old style" TODO will help simplify things.
https://github.com/Admiral-Piett/goaws/blob/master/app/utils/utils.go#L95

app/models/responses.go
```diff
@@ -80,20 +81,22 @@ func (r *ResultMessage) MarshalJSON() ([]byte, error) {
...
 
 	return json.Marshal(m)
 }
 
 type ResultMessageAttributeValue struct {
-	DataType    string `xml:"DataType,omitempty"`
-	StringValue string `xml:"StringValue,omitempty"`
-	BinaryValue string `xml:"BinaryValue,omitempty"`
+	DataType    string  `xml:"DataType,omitempty"`
+	StringValue *string `xml:"StringValue,omitempty"`
```
kojisaikiAtSony, line 98: Sorry to miss the description of this change.

`string` can't have "nil" value and it generates `"StringValue": ""` definitely on JSON serialize.
AWS SDK for Java won't calculate "BinaryValue" if "StringValue" is not null on their md5 hash validation, so the validation will be failed definitely if a message attribute has "BinaryValue".

app/models/responses.go
Admiral-Piett: Ok, I'm digging into this now a little more thoroughly as a result of this issue - https://github.com/Admiral-Piett/goaws/issues/328.  I think I understand what is going on in here now a little more.  

I think the overriding of the `MarshalJSON` method this way ties our hands, and there's a lot of different `"ResultMessageAttribute" VS "ResultAttribute"` stuff that makes this hard to consume.  I'm going to work on tweaking this a little and seeing if I can bring this piece more in line with the other end points.  Stand by!

Admiral-Piett: Ok, ok - I take it back.  🙃 .  The message attributes are a real mess and our flexibility is all tied up in this pattern, and to fix my issue I'll have to rebuild it all.  

If you, want I'll merge this unblock you if you're ready with all this, but I'm going to refactor the Message Attributes stuff now.  That will be my next order of business.  So after I do that, I'd appreciate it if you could smoke test it to make sure it still works for your team?

Thoughts on that plan?

Admiral-Piett (approved):

Admiral-Piett:
If you agree with all the above give us a rebase and we'll merge this?  Then I think we can let this be the last thing to go into the `0.5.1` release.  I'll do my refactoring after, so we have something to roll back to.  Thanks again!

app/models/responses.go
kojisaikiAtSony: OK! We'd be happy to apply this change to resolve our block. 
And of course, let me help you on the test with your refactoring on MessageAttributes! 💪  We can use our app and AWS SDK for Java.

kojisaikiAtSony:
Thank you for your accepting! We'll rebase soon 👍

app/models/responses.go
kojisaikiAtSony: As far as I look into #328, this PR won't resolve the issue with only itself. But it is related with jackson-ize process with empty map in go 🤔

kojisaikiAtSony:
Hi @Admiral-Piett , we rebased the branch and the tests passed in our local but Github action test looks failed. Could you please check it and merge this if there was no problem?
image

kojisaikiAtSony:
> @kojisaikiAtSony are you running your tests with a command like below? Make sure you try it with the `-race` flag.
> 
> ```
> go test -race ./app/...
> ```

@Admiral-Piett 
We tried `go test -cover -race ./app/...` that is same with the github action but these were succeeded...
We are using the below environments:
1. macOS sonoma + `go version go1.22.0 darwin/arm64`
2. Ubuntu 20.04.2 + `go version go1.23.2 linux/amd64`

Do you have any thoughts on the difference with the Github Action? And, how is it on your local?

Admiral-Piett:
@kojisaikiAtSony Sorry, I haven't had time to look into your test failings, but I will try to do that soon.  As an alternative, and like I think we talked about, I finished what I think is a pretty good refactor on the MessageAttributes.  I made a comment on PR for you to look at here - https://github.com/Admiral-Piett/goaws/pull/334/files#r1825116530, since I think that piece is the chunk that will really solve your issue.

Essentially it gets us out of the business of needing it encoded or decoded and leaves all that to marshaller.  It's pretty slick once we get the right type in place, and it just works.  I also kind of flipped your MarshalJSON logic, to make the "special" handling for the XML only, hoping I can delete that some day.

Let me know what you think about any or all of it!  I'm happy to discuss.  I can also build it for you if you want to test it before I merge it (and the model refactoring - they need to go together).  Either way is fine with me, but if we don't decide to put it in v0.5.1 it'll definitely be in v0.5.2, so you'll get this sooner or later.  :)

kojisaikiAtSony:
@Admiral-Piett I confirmed #334 will be an alternative to this PR, so I will close this PR later.
But I'd like to recreate a different PR that has the tests in this PR that asserts the hash. That would be required to keep the hash behavior since AWS SDK Go does not perform the hash validation.

Admiral-Piett:
@kojisaikiAtSony Sounds good, if I missed some tests by all means let me know what you're thinking, or if you have tests handy I'm happy to include those too.  Whatever works for you.  Thanks again for working with me on all this!

Admiral-Piett:
@kojisaikiAtSony 0.5.1 is out!  Thanks again for all your help!

kojisaikiAtSony:
I will add some tests for hash validation in separate PR.

PR #338, opened 2024-11-06
Implement message attribute parsing for PublishBatch

toastwaffle:
Factor out attribute parsing for SendMessage, SendMessageBatch, and Publish in the process.

I believe this change is adequately covered by the smoke tests, but am happy to be convinced otherwise

PR #339, opened 2024-11-13
Bugfix around message attributes on Publish(SNS), ReceiveMessages(SQS)

kojisaikiAtSony:
We found 3 bugs around message attributes, as shown below:
1. Received Message Attribute via XML is not decoded from base 64.
2. DataType value was broken by `caser` unexpectedly.
    * The exact conditions have not been identified, but at a certain point when publish/receive messages are repeated, the DataType value `String` becomes a string like `\u0000\u0000\u0000\u0000\u0000\u0000` due to the `caser`.
3. `MD5OfMessageAttributes` is missing on the message with message attributes from SQS-ReceiveMessages.

### DIFF
--- a/app/models/requests.go
+++ b/app/models/requests.go
@@ -1,6 +1,7 @@
 package models
 
 import (
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"net/url"
@@ -12,8 +13,6 @@ import (
 
 	log "github.com/sirupsen/logrus"
 )
-
-var caser = cases.Title(language.AmericanEnglish)
 
 type CreateQueueRequest struct {
 	QueueName  string            `json:"QueueName" schema:"QueueName"`
@@ -205,12 +204,18 @@ func (r *SendMessageRequest) SetAttributesFromForm(values url.Values) {
 		}
 
 		stringValue := values.Get(fmt.Sprintf("MessageAttribute.%d.Value.StringValue", i))
-		binaryValue := values.Get(fmt.Sprintf("MessageAttribute.%d.Value.BinaryValue", i))
+		encodedBinaryValue := values.Get(fmt.Sprintf("MessageAttribute.%d.Value.BinaryValue", i))
+
+		binaryValue, err := base64.StdEncoding.DecodeString(encodedBinaryValue)
+		if err != nil {
+			log.Warnf("Failed to base64 decode. %s may not have been base64 encoded.", encodedBinaryValue)
+			continue
+		}
 
 		r.MessageAttributes[name] = MessageAttribute{
 			DataType:    dataType,
 			StringValue: stringValue,
-			BinaryValue: []byte(binaryValue),
+			BinaryValue: binaryValue,
 		}
 	}
 }
@@ -257,16 +262,22 @@ func (r *SendMessageBatchRequest) SetAttributesFromForm(values url.Values) {
 		}
 
 		stringValue := values.Get(fmt.Sprintf("Entries.%d.MessageAttributes.%d.Value.StringValue", entryIndex, attributeIndex))
-		binaryValue := values.Get(fmt.Sprintf("Entries.%d.MessageAttributes.%d.Value.BinaryValue", entryIndex, attributeIndex))
+		encodedBinaryValue := values.Get(fmt.Sprintf("Entries.%d.MessageAttributes.%d.Value.BinaryValue", entryIndex, attributeIndex))
 
 		if r.Entries[entryIndex].MessageAttributes == nil {
 			r.Entries[entryIndex].MessageAttributes = make(map[string]MessageAttribute)
 		}
 
+		binaryValue, err := base64.StdEncoding.DecodeString(encodedBinaryValue)
+		if err != nil {
+			log.Warnf("Failed to base64 decode. %s may not have been base64 encoded.", encodedBinaryValue)
+			continue
+		}
+
 		r.Entries[entryIndex].MessageAttributes[name] = MessageAttribute{
 			DataType:    dataType,
 			StringValue: stringValue,
-			BinaryValue: []byte(binaryValue),
+			BinaryValue: binaryValue,
 		}
 
 		if _, ok := r.Entries[entryIndex].MessageAttributes[name]; !ok {
@@ -741,15 +752,21 @@ func (r *PublishRequest) SetAttributesFromForm(values url.Values) {
 		}
 
 		stringValue := values.Get(fmt.Sprintf("MessageAttributes.entry.%d.Value.StringValue", i))
-		binaryValue := values.Get(fmt.Sprintf("MessageAttributes.entry.%d.Value.BinaryValue", i))
+		encodedBinaryValue := values.Get(fmt.Sprintf("MessageAttributes.entry.%d.Value.BinaryValue", i))
+		binaryValue, err := base64.StdEncoding.DecodeString(encodedBinaryValue)
+		if err != nil {
+			log.Warnf("Failed to base64 decode. %s may not have been base64 encoded.", encodedBinaryValue)
+			continue
+		}
 
 		if r.MessageAttributes == nil {
 			r.MessageAttributes = make(map[string]MessageAttribute)
 		}
+
 		attributes[name] = MessageAttribute{
-			DataType:    caser.String(dataType), // capitalize
+			DataType:    cases.Title(language.AmericanEnglish).String(dataType), // capitalize
 			StringValue: stringValue,
-			BinaryValue: []byte(binaryValue),
+			BinaryValue: binaryValue,
 		}
 	}
 	r.MessageAttributes = attributes
--- a/app/models/requests_test.go
+++ b/app/models/requests_test.go
@@ -314,7 +314,7 @@ func TestSendMessageRequest_SetAttributesFromForm_success(t *testing.T) {
 	attr2 := r.MessageAttributes["Attr2"]
 	assert.Equal(t, "Binary", attr2.DataType)
 	assert.Empty(t, attr2.StringValue)
-	assert.Equal(t, []uint8("VmFsdWUy"), attr2.BinaryValue)
+	assert.Equal(t, []uint8("Value2"), attr2.BinaryValue)
 }
 
 func TestSetQueueAttributesRequest_SetAttributesFromForm_success(t *testing.T) {
--- a/app/models/responses.go
+++ b/app/models/responses.go
@@ -1,6 +1,7 @@
 package models
 
 import (
+	"encoding/base64"
 	"encoding/xml"
 )
 
@@ -83,6 +84,9 @@ func (r *ResultMessage) MarshalXML(e *xml.Encoder, start xml.StartElement) error
 	}
 	var messageAttrs []MessageAttributes
 	for key, value := range r.MessageAttributes {
+		if value.DataType == "Binary" {
+			value.BinaryValue = []byte(base64.StdEncoding.EncodeToString(value.BinaryValue))
+		}
 		attribute := MessageAttributes{
 			Name:  key,
 			Value: value,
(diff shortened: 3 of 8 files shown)


app/models/requests.go
```diff
@@ -205,12 +204,18 @@ func (r *SendMessageRequest) SetAttributesFromForm(values url.Values) {
...
+		if err != nil {
+			log.Warnf("Failed to base64 decode. %s may not have been base64 encoded.", encodedBinaryValue)
+			continue
+		}
 
 		r.MessageAttributes[name] = MessageAttribute{
 			DataType:    dataType,
 			StringValue: stringValue,
-			BinaryValue: []byte(binaryValue),
+			BinaryValue: binaryValue,
```
kojisaikiAtSony, line 218: This fixes Bug 1 "Received Message Attribute via XML is not decoded from base 64".
JSON requests has decoded from base 64 on JSON marshaller, but for XML, here `SetAttributesFromForm` does not have decoding step. So internal `BinaryValue` was still base 64 string (as binary) value.

app/models/requests_test.go
```diff
@@ -314,7 +314,7 @@ func TestSendMessageRequest_SetAttributesFromForm_success(t *testing.T) {
 	attr2 := r.MessageAttributes["Attr2"]
 	assert.Equal(t, "Binary", attr2.DataType)
 	assert.Empty(t, attr2.StringValue)
-	assert.Equal(t, []uint8("VmFsdWUy"), attr2.BinaryValue)
+	assert.Equal(t, []uint8("Value2"), attr2.BinaryValue)
```
kojisaikiAtSony, line 317: Internal value should be raw (not base 64 encoded) value.

app/models/responses.go
```diff
@@ -83,6 +84,9 @@ func (r *ResultMessage) MarshalXML(e *xml.Encoder, start xml.StartElement) error
 	}
 	var messageAttrs []MessageAttributes
 	for key, value := range r.MessageAttributes {
+		if value.DataType == "Binary" {
+			value.BinaryValue = []byte(base64.StdEncoding.EncodeToString(value.BinaryValue))
+		}
```
kojisaikiAtSony, line 89: Here is the related fix for Bug 1.

In the AWS query protocol, [blobs are sent and received after being base64 encoded](https://smithy.io/2.0/aws/protocols/aws-query-protocol.html) , so the binary value should be base64 encoded.

app/models/responses.go
```diff
@@ -95,6 +99,9 @@ func (r *ResultMessage) MarshalXML(e *xml.Encoder, start xml.StartElement) error
 	e.EncodeElement(r.MessageId, xml.StartElement{Name: xml.Name{Local: "MessageId"}})
 	e.EncodeElement(r.ReceiptHandle, xml.StartElement{Name: xml.Name{Local: "ReceiptHandle"}})
 	e.EncodeElement(r.MD5OfBody, xml.StartElement{Name: xml.Name{Local: "MD5OfBody"}})
+	if r.MessageAttributes != nil {
+		e.EncodeElement(r.MD5OfMessageAttributes, xml.StartElement{Name: xml.Name{Local: "MD5OfMessageAttributes"}})
+	}
```
kojisaikiAtSony, line 104: Here is the fix for Bug 3 "MD5OfMessageAttributes is missing on the message with message attributes from SQS-ReceiveMessages".

app/models/responses_test.go
```diff
@@ -63,7 +63,7 @@ func Test_ResultMessage_MarshalXML_success_with_attributes(t *testing.T) {
 	resultString := string(result)
 
 	// We have to assert piecemeal like this, the maps go into their lists unordered, which will randomly break this.
-	entry := "<ResultMessage><MessageId>message-id</MessageId><ReceiptHandle>receipt-handle</ReceiptHandle><MD5OfBody>body-md5</MD5OfBody><Body>message-body</Body>"
+	entry := "<ResultMessage><MessageId>message-id</MessageId><ReceiptHandle>receipt-handle</ReceiptHandle><MD5OfBody>body-md5</MD5OfBody><MD5OfMessageAttributes>message-attrs-md5</MD5OfMessageAttributes><Body>message-body</Body>"
```
kojisaikiAtSony, line 66: Assertion update for Bug fix 3.

app/models/responses_test.go
```diff
@@ -81,7 +81,7 @@ func Test_ResultMessage_MarshalXML_success_with_attributes(t *testing.T) {
 	entry = "<MessageAttribute><Name>attr1</Name><Value><DataType>String</DataType><StringValue>string-value</StringValue></Value></MessageAttribute>"
 	assert.Contains(t, resultString, entry)
 
-	entry = "<MessageAttribute><Name>attr2</Name><Value><BinaryValue>binary-value</BinaryValue><DataType>Binary</DataType></Value></MessageAttribute>"
+	entry = "<MessageAttribute><Name>attr2</Name><Value><BinaryValue>YmluYXJ5LXZhbHVl</BinaryValue><DataType>Binary</DataType></Value></MessageAttribute>"
```
kojisaikiAtSony, line 84: Assertion update for [blobs are sent and received after being base64 encoded](https://smithy.io/2.0/aws/protocols/aws-query-protocol.html).

smoke_tests/sns_publish_test.go
```diff
@@ -64,7 +65,92 @@ func Test_Publish_sqs_json_raw(t *testing.T) {
...
+		Protocol:              aws.String("sqs"),
+		TopicArn:              createTopicResult.TopicArn,
+		Attributes:            map[string]string{},
+		Endpoint:              createQueueResult.QueueUrl,
+		ReturnSubscriptionArn: true,
+	})
+
+	snsClient.SetSubscriptionAttributes(context.TODO(), &sns.SetSubscriptionAttributesInput{
+		SubscriptionArn: subscribeResult.SubscriptionArn,
+		AttributeName:   aws.String("RawMessageDelivery"),
```
kojisaikiAtSony, line 100: This new test pattern is for "RawMessageDelivery".

smoke_tests/sqs_receive_message_test.go
```diff
@@ -268,7 +268,7 @@ func Test_ReceiveMessageV1_xml_with_attributes(t *testing.T) {
 	assert.Equal(t, 1, len(receiveMessageResponse.Result.Messages))
 	assert.Equal(t, "MyTestMessage", receiveMessageResponse.Result.Messages[0].Body)
 	assert.Equal(t, "ad4883a84ad41c79aa3a373698c0d4e9", receiveMessageResponse.Result.Messages[0].MD5OfBody)
-	assert.Equal(t, "", receiveMessageResponse.Result.Messages[0].MD5OfMessageAttributes)
+	assert.Equal(t, "ae8770938aee44bc548cf65ac377e3bf", receiveMessageResponse.Result.Messages[0].MD5OfMessageAttributes)
```
kojisaikiAtSony, line 271: xml/json both should have hash for message attribute if the message attributes is not empty.

smoke_tests/sqs_send_message_batch_test.go
```diff
@@ -370,7 +372,7 @@ func TestSendMessageBatchV1_Xml_Success_including_attributes(t *testing.T) {
 		WithFormField("Entries.1.MessageBody", messageBody2).
 		WithFormField("Entries.1.MessageAttributes.1.Name", binaryAttributeKey).
 		WithFormField("Entries.1.MessageAttributes.1.Value.DataType", binaryType).
-		WithFormField("Entries.1.MessageAttributes.1.Value.BinaryValue", binaryValue).
+		WithFormField("Entries.1.MessageAttributes.1.Value.BinaryValue", binaryValueEncodeString).
```
kojisaikiAtSony, line 375: Here is also for [blobs are sent and received after being base64 encoded](https://smithy.io/2.0/aws/protocols/aws-query-protocol.html).
BinaryValue in XML should be base 64 encoded.

app/models/requests.go
```diff
@@ -741,15 +752,21 @@ func (r *PublishRequest) SetAttributesFromForm(values url.Values) {
...
+			continue
+		}
 
 		if r.MessageAttributes == nil {
 			r.MessageAttributes = make(map[string]MessageAttribute)
 		}
+
 		attributes[name] = MessageAttribute{
-			DataType:    caser.String(dataType), // capitalize
+			DataType:    cases.Title(language.AmericanEnglish).String(dataType), // capitalize
```
kojisaikiAtSony, line 767: Here is the fix for Bug 2 "DataType value was broken by `caser` unexpectedly".

As I wrote in the description, I don't know the exact conditions under which this occurs, but since there was no need to make it a global variable, I changed it to a local definition and the problem was resolved. If you have any ideas, I would appreciate your comments.

app/models/requests.go
Admiral-Piett: I don't really love this.  I know it's more true to life, since that input is supposed to be encoded, but I'd prefer to make the decoding optional.  It's a testing tool after all, and we're doing our best to cover the needs of all.  It's not like we're throwing errors for invalid attributes or anything.

So maybe instead of your continue, the log is fine, but take the value as is and stuff it in the resulting attribute?

app/models/responses.go
Admiral-Piett: I don't know about this.  I understand that's part of the query protocol, but that whole thing is deprecated in AWS.  Based on my tests this works, as is, for the existing sdks I've tried.  What exactly isn't working for you?

app/models/responses.go
Admiral-Piett: I'm not sure why I left this out.  I think there was a reason though.  A lot of this had to do with the raw vs regular message handling.  Can you tell me more about what your issues are on this?

Admiral-Piett (commented):
I need some more information on this one before we move forward.  Having just done some significant work on this, I want to understand more about what's causing you problems before we change it.  Also, I think there are a couple opportunities to be the "path of least resistance" here, and I think that's helpful.  See my individualized comments for the specifics.

Finally - https://github.com/Admiral-Piett/goaws/pull/338 @toastwaffle did some good work on combining the logic around parsing incoming MessageAttributes.  It'll definitely affect your change, since you'd only need to do it once.  Want to take a look at what they've got first?  I think there's is pretty close and I think it makes sense to get that in before this, so we can just fix the bugs in 1 area, if you're ok with that.

Thanks for looking at this, as always!

app/models/requests.go
kojisaikiAtSony: OK! I will update it as the below 👍 
* Try decoding
    * Success: store it as the decoded raw value
    * Failure: log and store it as original requested value

app/models/responses.go
kojisaikiAtSony: Yep, as you said this part is for the query protocol. The latest version of AWS SDK for Java also works without this part.

I thought that future versions of GOAWS still keep support the query protocol, so I added it.
Sorry I don't remember clearly but, Is the policy that "users who need to use aws-query should use older versions of goaws"? (It also makes sense)

app/models/requests.go
kojisaikiAtSony: Yep, I thought you said so 😅. It's a little creepy when the behavior isn't clear. Let's dive into the caser code and take a closer look 👍

kojisaikiAtSony:
> Finally - https://github.com/Admiral-Piett/goaws/pull/338 @toastwaffle did some good work on combining the logic around parsing incoming MessageAttributes.
The PR is very helpful! We'll rebase this once it is merged 👍

app/models/requests.go
Admiral-Piett: Sorry @kojisaikiAtSony - I'm looking at this again, and more of my previous thoughts are coming back to me.  Looks to me like this change is decoding it, storing it internally, and then upon receiving the message, encoding it, and returning it.  Double check me, is that what you're doing here?

If so, why do this at all?  That's why I undid all this decoding/encoding business not long ago.  It didn't seem like there was a point.  Are you having trouble with the checksum or something because it's encoded?   Where are you seeing the "Received Message Attribute via XML is not decoded from base 64" error, and how are you recreating it?

app/models/responses.go
Admiral-Piett: Well, no, that's not the policy per say.  I hadn't given it that much thought if I'm honest.  But if you think about it like this - this is a tool for folks to run their own apps against instead AWS directly.  So my main goal is to make it work in a performant/lightweight way and replicate as much of the AWS SQS/SNS "contract" as I can.  

So with that in mind - we don't NEED to match what AWS does, as long as we respond the same way that it would.  We just need to match their contracts right?  At any rate we can't know what they're doing all the times as their docs are generally garbage in my opinion.  

So, I don't care about things like security, and base64 encoding much, because, it's really up to the user here to worry about their input and output - because it's an app running where a user puts it.  "Do what you will with it." so to speak, as long as an SDK doesn't fail on it - "garbage in garbage" out.  

**Here's what I need:**
So all that said - this change is doing a couple of things.  I need to know why we need them.  The way I had it lately was working for me, based on the contract I had researched.  What I'd like to see is what you're failing with, how you're recreating it, and why we think these changes will fix them.  

Sorry - this is a lot for me to juggle long term and I need a little grooming/refining (in agile terms), before I let it in.  Otherwise I'll be adrift if someone else complains about it or something similar later.

As always thank you for giving it so much work and thought.  I really appreciate it and without you I wouldn't be able to keep this tool as current and viable as it is.

kojisaikiAtSony:
@Admiral-Piett 
I'm really sorry for creating such a big PR without considering your burden 🙇 
I will create separated, small PRs on each fixes.

(One of the reasons I'm focusing on this improvement is to silence security teams who keep telling us to update the old AWS SDKs... 😅)

PR #345, opened 2025-01-23
Set `VisibilityTimeout` on SQS messages to the value included in the `ReceiveMessage` request

cainenielsen:
Here I am making a change to the `ReceiveMessage` request handler wherein, if the `requestBody` contains a `VisibilityTimeout` property it will be set on the message instead of defaulting to the related queues' default `VisibilityTimeout` value.

### DIFF
--- a/app/gosqs/receive_message.go
+++ b/app/gosqs/receive_message.go
@@ -108,7 +108,12 @@ func ReceiveMessageV1(req *http.Request) (int, interfaces.AbstractResponseBody) 
 			}
 			msg.ReceiptHandle = msg.Uuid + "#" + randomId
 			msg.ReceiptTime = time.Now().UTC()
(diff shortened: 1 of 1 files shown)


cainenielsen:
Some documentation surrounding the `ReceiveMessage` operation

- [AWS SDK (GoLang V2) ReceiveMessageInput type](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/sqs#ReceiveMessageInput)
- [AWSAPI Documenation - SQS ReceiveMessage Request Parameters](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_ReceiveMessage.html#API_ReceiveMessage_RequestSyntax)

Admiral-Piett:
@cainenielsen Fair enough, this looks good.  Thanks for looking into it!  Could you add a unit test?  After that I think we can let it rip.  Thanks!

Admiral-Piett (approved):
Lovely, thank you!

